PR #1, opened 2026-05-12
feat(lang): add Clojure language support

everapp-org:
## Summary

- Adds Clojure as the 43rd supported language via tree-sitter-clojure 0.1.0
- Extracts ns, defn/defn-, defmacro, def, defprotocol, defrecord, deftype, defmulti definitions
- Handles ^:private metadata for visibility detection
- Qualifies symbols as my.ns/fn-name via two-pass extraction
- Parses :require/:use clauses for import tracking
- Maps .clj, .cljs, .cljc, .edn extensions to the clojure language
- Adds semicolon comment prefix for context annotation
- Adds clojure to cross-language HTTP/IPC capability list
- 16 unit tests covering all definition forms, visibility, imports, and line numbers

## Test plan

- [ ] cargo test --features lang-clojure — all 16 Clojure unit tests pass
- [ ] cargo test — full suite passes, language count assertion >= 43 satisfied
- [ ] cargo build — compiles cleanly with default features


src/parser/languages/clojure.rs
```diff
@@ -0,0 +1,717 @@
...
+
+    #[test]
+    fn test_private_def_via_metadata() {
+        let src = "(ns my.ns)\n(def ^:private secret 42)\n";
+        let result = parse_and_extract(src);
+        let vars: Vec<_> = result
+            .symbols
+            .iter()
+            .filter(|s| s.name == "my.ns/secret")
+            .collect();
```
lyubomir-bozhinov, line 646: Nice test for the `def` form. The `defn` arm a few hundred lines above also calls `list_name_text` and checks `is_meta_private`, but no test exercises that path on a `defn`. A refactor that broke metadata detection on `defn` specifically (without touching `def`) would slip through silently.

Could you add a mirror test? Something like:

```rust
#[test]
fn test_private_defn_via_metadata() {
    let src = "(ns my.ns)\n(defn ^:private hidden [] :ok)\n";
    let result = parse_and_extract(src);
    let fns: Vec<_> = result
        .symbols
        .iter()
        .filter(|s| s.name == "my.ns/hidden")
        .collect();
    assert!(!fns.is_empty(), "expected ^:private defn 'hidden'");
    assert_eq!(fns[0].visibility, Visibility::Private);
    assert!(
        !result.exports.iter().any(|e| e.name == "my.ns/hidden"),
        "^:private defn should not be exported"
    );
}
```

That closes the only meaningful coverage gap relative to the PR summary's claim.

src/parser/languages/clojure.rs
```diff
@@ -0,0 +1,717 @@
...
+            result.symbols.iter().any(|s| s.name == "standalone"),
+            "unqualified defn should have plain name"
+        );
+    }
+
+    #[test]
+    fn test_multiple_definitions() {
+        let src = r#"
+(ns my.app
+  (:require [clojure.string :as str]))
```
lyubomir-bozhinov, line 701: Four of these assertions check only `s.kind == X` — so a regression that broke `defprotocol` / `defrecord` / `defmacro` / `ns` extraction by emitting the wrong name (or wrong namespace qualifier) would still pass this test. Since the fixture has the right symbols, tightening each to `kind && name` is mechanical:

```rust
assert!(result.symbols.iter().any(|s| s.kind == SymbolKind::Class && s.name == "my.app"));
assert!(result.symbols.iter().any(|s| s.kind == SymbolKind::Constant && s.name == "my.app/version"));
assert!(result.symbols.iter().any(|s|
    s.kind == SymbolKind::Function
        && s.visibility == Visibility::Private
        && s.name == "my.app/build-url"
));
assert!(result.symbols.iter().any(|s| s.kind == SymbolKind::Function && s.name == "my.app/fetch"));
assert!(result.symbols.iter().any(|s| s.kind == SymbolKind::Interface && s.name == "my.app/Repository"));
assert!(result.symbols.iter().any(|s| s.kind == SymbolKind::Struct && s.name == "my.app/InMemoryRepo"));
assert!(result.symbols.iter().any(|s| s.kind == SymbolKind::Macro && s.name == "my.app/with-retry"));
assert!(result.imports.iter().any(|i| i.source == "clojure.string"));
```

Each assertion now individually proves its arm of `process_top_level` produces the right qualified name, not just *some* symbol of the right kind.

lyubomir-bozhinov (changes requested):

lyubomir-bozhinov:
Thanks for this — really thoughtful submission. The 707-line parser handles Clojure's idioms correctly (the `meta_lit` traversal for `^:private`, the two-pass ns-qualification, both `[ns :as alias]` vector style and bare-symbol require), the trait implementation is faithful to the sibling parsers, and the security audit came back clean: zero panic vectors in production code, zero `unsafe`, no extension collisions, fault-tolerant `unwrap_or` / explicit `None` handling throughout. The 16 tests cover every form claimed in the summary except for two gaps I've called out inline.

## Before merging

### Doc-drift in 4 files — "tree-sitter across 42 languages" wasn't bumped to 43

These files aren't in the PR diff so I can't anchor inline, but each needs a `42` → `43` bump:

- [ ] **`README.md`** — line 11 (`tree-sitter across 42 languages`), line 176 (`## Language support (42)`), and the Tier-1 list at ~179 needs `Clojure` appended.
- [ ] **`.claude/CLAUDE.md`** — line 143 (`## Supported Languages (42)`) plus the Tier-1 list at 145–147 needs `Clojure` appended. CLAUDE.md is loaded into every `cxpak_auto_context` call so keeping the count accurate has compounding downstream effect.
- [ ] **`plugin/.claude-plugin/plugin.json`** — line 3 description text contains `tree-sitter across 42 languages`.
- [ ] **`.claude-plugin/marketplace.json`** — line 10 description is character-identical to plugin.json. Same bump.

### Tests — two real coverage gaps

- [ ] Add a `test_private_defn_via_metadata` (mirroring `test_private_def_via_metadata`). See inline comment.
- [ ] Tighten the four kind-only assertions in `test_multiple_definitions`. See inline comment.

## Nice-to-have, defer to follow-up if you prefer

None of these block merge — happy to take them as separate PRs, or address them here if it's convenient:

- `intelligence/api_surface.rs::is_source_code_file` (~line 299-334) doesn't list `clj/cljs/cljc/edn` — you added Clojure to `is_web_code_language`, so route detection won't fire on `.clj` files. Caveat: lua/haskell also absent — pre-existing patchy pattern, not introduced by this PR.
- `parser/mod.rs:188` registry-floor was bumped to `>= 43` unconditionally — same shape as the prior `>= 42` so not a new fragility, just noting that curated `--no-default-features --features <subset>` builds now fail on the new floor.
- `test_symbol_line_numbers` asserts only `start_line == 3`; `end_line` correctness on multi-line bodies is untested.
- `deftype` shares the `defrecord` branch but has no dedicated test (only the kind-only check from `test_multiple_definitions`).
- `:refer [foo bar]` import vector path is unexercised — `test_require_imports_extracted` uses `:as` only.
- Reader conditionals `#?(:clj … :cljs …)` (common in `.cljc`), multi-arity `(defn f ([x] …) ([x y] …))`, comment-only files, and malformed source aren't tested.
- The `if ns.is_empty() { name.clone() } else { format!("{}/{}", ns, name) }` pattern is duplicated identically in 7 match arms — extracting a `qualify(ns, name)` helper would tighten this.
- `named_children()` returns `Vec<Node>`; a cursor iterator would avoid the per-call allocation. Pure idiom nit.
- `:use` is advertised in the PR summary but has thinner inline docs than `:require`.

## For the record (not blocking)

Identifier sanitization (bidi / control-char / length-cap) on extracted symbol names is a **repo-wide gap** that affects all 42 existing tier-1 parsers identically — cxpak handles bidi at the presentation layer (`search_index.rs`, `architecture.rs`, `lsp/methods.rs`), not at extraction. I'll file a separate tracking issue covering all parsers. This PR shouldn't be held up for an architectural decision that predates it.

---

Approval follows as soon as the doc-drift items + two test additions are addressed. Thanks for the careful work — Clojure was one of the most-requested missing languages.

lyubomir-bozhinov:
Pushed `bac714c` to drop the `.claude/CLAUDE.md` file and the `!.claude/CLAUDE.md` exception in `.gitignore`. That was on me — my review referenced `.claude/CLAUDE.md` as needing a 42 → 43 bump, but that path is a maintainer's user-local file, not a repo artifact. Sorry for the misdirection.

The doc-drift fixes in `README.md`, `plugin.json`, and `marketplace.json` are exactly right and stay. Tests and other changes look great — once CI on the new HEAD is green I'll approve.

lyubomir-bozhinov (approved):
LGTM — all CI green.

Two rounds of follow-up commits and you've cleaned up every blocking item plus the entire nice-to-have list I'd flagged for follow-up PRs.

**Round 1 (`bacc30e`)** — doc-drift 42 → 43 across README + plugin.json + marketplace.json; `test_private_defn_via_metadata` mirroring the `def` test exactly as drafted; `test_multiple_definitions` tightened with `kind && name` assertions on all 8 arms.

**Round 2 (`7249be9`)** — `is_source_code_file` now recognises `.clj/.cljs/.cljc/.edn`; the seven duplicated `qualify(ns, name)` inline blocks collapsed into one helper; `named_children()` iterator replaces the per-call allocation; registry-floor canary now carries a comment explaining its unconditional nature; `extract_ns_imports` docs cover `:use` and `:refer`; and six new tests close the remaining surface gaps — `deftype` as `Struct`, `:refer` import vectors, reader conditionals not panicking, multi-arity `defn`, comment-only files, and malformed source. `end_line` correctness is now asserted on multi-line bodies too.

23 tests total covering every form the parser claims to handle, zero `unsafe`, fault-tolerant `unwrap_or` throughout, two-pass ns qualification working correctly on every arm of `process_top_level`. Thanks for the careful work — Clojure was one of the most-requested missing languages, and this lands at a quality bar I'd be happy to see other language additions match.

PR #3, opened 2026-06-10
docker: add Dockerfile, standalone image, and usage docs

everapp-org:
## Summary

- **`Dockerfile`** — multi-stage source build; deps cached in a separate layer to keep rebuilds fast
- **`Dockerfile.standalone`** — fetches the pre-built upstream binary from GitHub Releases; no source tree needed, works on amd64 and arm64; uses `ubuntu:24.04` to match the glibc version of the upstream release builder
- **`.dockerignore`** — excludes `target/`, `.git/`, docs, and tests from the build context
- **README** — new Docker section covering standalone vs source builds, macOS/Linux and Windows (PowerShell) usage, HTTP server auth requirements, and MCP transport/scope clarification

## Key behaviours documented

- `--bind 0.0.0.0 --token <secret>` is required for the HTTP server to be reachable from the host; cxpak enforces this at startup
- MCP runs over stdio (JSON-RPC), not HTTP — one instance per repo
- On Windows, use `curl.exe` — PowerShell aliases `curl` to `Invoke-WebRequest` which fails silently against this server

## Motivation

Docker makes cxpak accessible on platforms where a native binary isn't available (Windows) and in environments where managing a Rust toolchain is undesirable (CI, sandboxed servers, air-gapped machines). The standalone image requires no source checkout — just the single `Dockerfile.standalone`.

## Test plan

- [ ] `docker build -f Dockerfile.standalone -t cxpak .` completes without error
- [ ] `docker run --rm -v $(pwd):/repo cxpak overview .` returns codebase summary
- [ ] `docker run -d -p 3000:3000 -v $(pwd):/repo -v cxpak-models:/root/.cxpak cxpak serve --bind 0.0.0.0 --token mysecret .` starts and `curl http://localhost:3000/health` responds
- [ ] Tested on Windows 10 with Docker Desktop (PowerShell)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lyubomir-bozhinov (changes requested):
## 🐳 Review: Docker support — **Request Changes**

Thanks for this — it's a genuinely clean, well-scoped addition. The two-Dockerfile split (source vs. standalone) is sensible, the docs are unusually accurate, and the change is atomic (4 files, +179/−0, no source touched). I verified the documented behaviour against the actual code and it all checks out:

- **Docs match reality** — default bind is `127.0.0.1` and `validate_bind_security()` (`src/commands/serve.rs:969`) does refuse a non-loopback bind without a token; `/health` is unauthenticated (`serve.rs:342`) while `/v1/*` requires Bearer; `--bind`/`--token`/`--mcp` all exist; default port is `3000`, matching `EXPOSE 3000`. The `curl.exe` PowerShell caveat is real and a nice touch.
- **`.dockerignore` is build-safe** — every `include_str!` target lives under `assets/` (which is `COPY`'d), and `build.rs`/`Cargo.toml` reference no excluded path. `daemon`/`embeddings`/`lsp`/`visual` are all *default* features, so `cargo build --release` produces a fully-featured binary (serve + MCP included). 👍
- **glibc reasoning is correct** (bookworm builder + bookworm runtime; `ubuntu:24.04` standalone matching the release builder), **apt hygiene is good** (`--no-install-recommends` + `rm -rf /var/lib/apt/lists/*` everywhere), the multi-stage dep-cache works, and there are no hardcoded secrets (`mysecret` is clearly a placeholder).

The blockers below are all supply-chain integrity — not functional defects.

### 🔴 Required before merge

**1. Pin every image and the downloaded binary by digest / checksum.**

All four base images are pinned to *mutable* tags, so builds aren't reproducible and a re-pushed upstream tag flows in silently:
- `Dockerfile:3` → `rust:1.91-slim-bookworm`
- `Dockerfile:28` → `debian:bookworm-slim`
- `Dockerfile.standalone:16` and `:36` → `ubuntu:24.04`

Please digest-pin (keep the tag as a readable comment):
```dockerfile
FROM debian:bookworm-slim@sha256:<digest>
```
Resolve with `docker buildx imagetools inspect <image>`. ⚠️ For `Dockerfile.standalone` use the **multi-arch manifest-list digest**, not a single-platform one, or you'll break the amd64/arm64 `TARGETARCH` logic.

And the release binary is fetched with **no integrity check** at `Dockerfile.standalone:31-33`:
```dockerfile
curl -fsSL ".../cxpak-${TRIPLE}.tar.gz" | tar -xz -C /usr/local/bin
```
HTTPS-from-your-own-releases bounds the risk, but the artifact bytes are unverified and GitHub release assets are mutable. This project already pins per-arch `sha256` for the Homebrew formula (`.github/workflows/release.yml:120-123`), so the Docker path should match that bar. Suggested shape — download to a file, verify, then extract (no `curl | tar` pipe):
```dockerfile
ARG VERSION
ARG SHA256_AMD64
ARG SHA256_ARM64
# select $EXPECTED_SHA per $TARGETARCH, then:
RUN curl -fsSL "$URL" -o cxpak.tar.gz \
 && echo "${EXPECTED_SHA}  cxpak.tar.gz" | sha256sum -c - \
 && tar -xzf cxpak.tar.gz -C /usr/local/bin && rm cxpak.tar.gz
```
Because the image is version- and arch-parameterized, a single hardcoded SHA won't do — either pass per-arch SHAs as build-args, or (cleaner long-term) publish a `SHA256SUMS` asset with each release and verify against it. Cosign/minisign signing would be the gold standard.

**2. Bump the stale default version.** `Dockerfile.standalone:13` is `ARG VERSION=2.1.0`, but the current release is **2.2.1**. A plain `docker build -f Dockerfile.standalone .` (no `--build-arg`) silently builds an image a full minor behind. Please set `ARG VERSION=2.2.1` (or drop the default and require it). The help example at `:6` (`VERSION=2.0.0`) is also worth refreshing.

### 🟡 Recommended (non-blocking)

- **Run as non-root.** Neither image sets a `USER`, and `VOLUME ["/root/.cxpak"]` (`Dockerfile:42`, `Dockerfile.standalone:46`) confirms the process runs as uid 0. Adding a `useradd` + `USER` and relocating the model-cache off `/root` (with matching README `-v` examples) limits blast radius if the server is ever exposed.
- **Rebase + refresh a now-stale comment.** This branch predates the just-merged #4, which removed OpenSSL from the tree (`git2 = default-features = false`, `reqwest → rustls-tls`). So the `Dockerfile:5` comment *"cmake + perl are required by vendored OpenSSL (git2 dep)"* is no longer accurate on `main`. Please rebase and re-verify which builder build-deps are actually needed against current `main`, then correct the comment. (Good news: the runtime stage stays correct either way — no system TLS library is needed in either the old vendored-static or the new rustls configuration.)

### 🧪 Test coverage

The PR test plan checkboxes are unticked, and the source-`Dockerfile` build path isn't exercised by CI (CI doesn't build images). A one-line confirmation that `docker build .` (source) and `docker build -f Dockerfile.standalone .` both succeed locally would close the loop.

---

**Summary:** Solid, accurate, well-scoped work — the only thing standing between this and a merge is supply-chain pinning (#1) and the stale version default (#2). Thanks again! 🙏

PR #23, opened 2026-07-16
fix(mcp): keep serve --mcp index fresh via an owned file watcher (#18)

### LINKED ISSUE #18: serve --mcp serves a stale index after file edits (HTTP serve watches; MCP path does not)
## Summary

`cxpak serve --mcp` builds its index **once at startup and never refreshes it**. After any working‑tree
edit, every structural query (`graph`/`call_graph`, `dead_code`, `security_surface`, `health`,
`architecture`, `verify`, `predict`, `context`, …) keeps returning results computed against the
**startup snapshot**, silently — no error, no staleness signal, and no way to force a refresh short of
restarting the process.

The HTTP server (`cxpak serve`, no `--mcp`) does **not** have this problem: it spawns a `FileWatcher` and
incrementally updates a swappable index on every file change. The MCP path is missing that wiring, so the
two transports have **asymmetric freshness semantics** for the same command.

Version: **cxpak 3.1.0** (`git 8b5c279`).

## Impact

Any long‑lived MCP consumer that keeps one session open across edits gets **stale structural answers**.
The concrete case that surfaced this: a persistent/warm `serve --mcp` session reused across a sequence of
edits (e.g. a commit‑gate that verifies each change) — by the second edit, `call_graph`/`dead_code`/
`security_surface` reflect the *original* tree, not the current one. Because it fails **silently** (stale
data, not an error), consumers can act on wrong results. Today the only correct pattern is
cold‑spawn‑per‑query (pay the full 10–14 s index every call), which defeats the point of a warm session.

## Root cause (traced in `src/commands/serve.rs`)

`run_mcp` does a **one‑shot** background build and then serves that fixed cell — there is no `FileWatcher`
and no incremental‑update loop:

- `run_mcp` — `src/commands/serve.rs:2238`
- `spawn_mcp_index_build(path, readiness)` — `serve.rs:2250` → a single `build_index` → publishes
  `IndexReadiness::Ready(Arc<CodebaseIndex>)` (`serve.rs:61`) **once**
- `mcp_stdio_loop` (`serve.rs:2450`) then answers every `tools/call` from that unchanging cell

Contrast the HTTP path, which **does** watch and hot‑update:

- `run` — `serve.rs:1351`
- `FileWatcher::new(...)` — `serve.rs:1384`
- `process_watcher_changes(&changes, &watcher_path, &watcher_index)` — `serve.rs:1397`, driving the
  atomically‑swappable `SharedIndex` via `commands::watch::{apply_incremental_update, classify_changes}`

So the freshness machinery already exists and is proven; it is simply **not wired into `run_mcp`**.

## Reproduction

1. In a repo, `cxpak serve --mcp .` — startup logs one line: `cxpak: MCP index ready (N files indexed, …)`.
2. From an MCP client, call a structural tool (e.g. `graph`/`call_graph`) and note a symbol/edge.
3. Edit a source file to add or remove a symbol (or introduce an unresolved call).
4. Call the same tool again on the **same** session.

**Expected:** the result reflects the edit (as the HTTP server would after its watcher re‑indexes).
**Actual:** the result is unchanged (startup snapshot); the server logs no re‑index, and there is no
`reindex`/`refresh` tool to force one. Only restarting `serve --mcp` picks up the change.

(The one‑shot build in `run_mcp` above is the authoritative evidence; the HTTP server's single startup
`… files indexed` log with no re‑index line on edit is the same behavior on that transport's build step.)

## Why the obvious workaround doesn't work

Running a separate `cxpak watch` alongside `serve --mcp` does **not** help: `run_mcp` holds the index **in
memory** (the `Ready` cell), it does not re‑read `.cxpak/` per query. A `watch` process keeping the
on‑disk index hot is therefore invisible to the running MCP server.

## Suggested fix (small — reuse the HTTP path's primitives)

Wire the existing watcher into `run_mcp`. After the initial build publishes `Ready(Arc<CodebaseIndex>)`,
spawn the **same** watcher thread the HTTP `run` uses (`FileWatcher` + `process_watcher_changes`) and, on
each change batch, apply `apply_incremental_update` to the current index and **republish**
`Ready(new_index)`. `mcp_stdio_loop` already snapshots the readiness cell per `tools/call`, so it would
transparently pick up the swapped index — no protocol change, no new subsystem, ~30–50 lines reusing
`daemon::watcher::FileWatcher`, `process_watcher_changes`, and `commands::watch::{apply_incremental_update,
classify_changes}`.

Nice‑to‑have alongside it: an explicit `reindex`/`refresh` MCP tool for clients that want to force a
rebuild deterministically (e.g. right after a commit) rather than rely on the debounce window.

## Environment

- cxpak 3.1.0 (`8b5c279`), macOS (Apple Silicon), installed via `barnett-studios/tap`.
- Transport: MCP stdio (`cxpak serve --mcp <path>`).

lyubomir-bozhinov:
Fixes #18.

## Problem
`cxpak serve --mcp` built its index once at startup and never refreshed it (its own doc comment admitted "no periodic rebuild"). A warm MCP session returned stale structural answers after any edit — silently, no signal, no way to refresh short of restarting. The HTTP server (`cxpak serve`) already watched and hot-updated; the two transports had asymmetric freshness.

## Fix — ADR-0200 (in this PR)
Wire the proven HTTP watcher primitives into `run_mcp`, but with an **owned, terminable** thread (the HTTP watcher is a detached infinite loop — fine for a process-lifetime server, but it would leak into in-process tests and can't be joined on shutdown):
- `spawn_mcp_watcher` waits for the background build to become queryable (`Ready` **or** `ReadyEnriched`), seeds a mirrored `SharedIndex`, runs the existing `process_watcher_changes` on each debounced edit, and republishes `Ready(new)` into the readiness cell — which `tools/call` already snapshots per call.
- Lifecycle is owned: a `shutdown` `AtomicBool` is checked in both the pre-Ready wait (every 100ms → stdin EOF during the 13–34s build joins promptly) and the main poll loop; a `Failed` build aborts the watcher; `FileWatcher::new` failure is fail-open. `run_mcp` sets `shutdown` and joins both threads on exit.

### The subtle part — a race this fix had to *close*, not introduce
The opt-in embedding-enrichment phase (`publish_ready_enriched`, ADR-0186) is a **second writer** of the readiness cell: it blind-swapped `ReadyEnriched(startup_base)` seconds after `Ready`. Left naive, a watcher republish landing in that window would be silently clobbered by the late enrichment — reintroducing #18's exact defect with an embedding-adorned stale index. Fix: `publish_ready_enriched` is now a **compare-and-swap** — it swaps `ReadyEnriched` only if the cell still holds the exact base `Arc` it enriched (`Arc::ptr_eq`), else skips + logs. The `RwLock::write()` guard spans the full check-then-set, so it's a true atomic CAS. Every interleaving converges to the fresh index. Each watcher republish also clears `embedding_index` (a delta-rebuilt clone would otherwise serve a stale 7th signal). (This race was caught by the validation gate before implementation, not in review.)

The `refresh`/`reindex` MCP tool (the issue's stated *nice-to-have*) is deferred — it would be a third, uncoordinated writer of the cell; the debounced watcher fixes the core bug.

## Before / after (manual QA, real MCP stdio path)
```
BEFORE edit  node(src/foo.rs): out_degree 0
< edit: a.rs gains an out-edge, git commit >
AFTER edit   node(src/foo.rs): out_degree 1   <-- WARM session, no restart
```
(The Phase-0 reproduction served `out_degree 0` after the edit; now it reflects it.)

## Tests (RED→GREEN, deterministic)
Pure-function unit tests (no fixed-sleep asserts): enrich CAS skips-when-moved / swaps-when-unchanged; `classify_watcher_wait` (Ready/ReadyEnriched/Building/Failed); `republish_watcher_index` clears embedding; shutdown-on-signal and shutdown-during-pre-Ready-wait; abort-on-failed-build. Plus a **real-subprocess integration test** (`tests/mcp_freshness_subprocess.rs`): real `serve --mcp` binary, real stdio JSON-RPC, real FS edit through the `notify` FileWatcher, asserting the warm session reflects the edit and exits cleanly on stdin EOF — all waits bounded.

- `cargo test --workspace --features daemon`: green (incl. embeddings-gated CAS tests). `clippy --all-targets --all-features -D warnings`: clean. `fmt --check`: clean.
- Golden fixture `spa_output_matches_golden_fixture`: byte-identical. HTTP `run` watcher path untouched.
- Independent pre-PR review: enrich race confirmed fully closed for all interleavings; no blockers.

### Known, accepted limitations
- A watcher swap drops embedding enrichment (7→6 signal) until a future re-enrich (out of scope; logged).
- The `FileWatcher::new`-failure fail-open branch has no cheap failure-injection seam (scoped to the CI 90% coverage floor).
- A panic unwinding through the stdio loop would skip the thread joins — a pre-existing pattern (the build thread had the same), not a regression.

https://claude.ai/code/session_019hJMLmPVoGS2FeUUX4w4Jz

### DIFF
--- a/docs/adrs/0200-mcp-index-freshness-via-watcher.md
+++ b/docs/adrs/0200-mcp-index-freshness-via-watcher.md
@@ -12,18 +12,18 @@ loop: planning
 ## Context
 
 `cxpak serve --mcp` builds its index **once** at startup (`run_mcp` →
-`spawn_mcp_index_build`, `src/commands/serve.rs:2238/2318`, ADR-0185) and never
-refreshes it. Its own doc comment admits it (serve.rs:2448-2449: "a long-lived
+`spawn_mcp_index_build`, ADR-0185) and never
+refreshes it. Its own doc comment admits it (`snapshot_ready_index`: "a long-lived
 session would keep the first ready index (no periodic rebuild)"). Reproduced
-(REPRO.md): a warm MCP session returns `out_degree 0` for a symbol after an edit
+empirically: a warm MCP session returns `out_degree 0` for a symbol after an edit
 that a fresh index shows as `out_degree 1` — silently, no staleness signal.
 
-The HTTP path (`run`, serve.rs:1351) does **not** have this bug: it spawns a
+The HTTP path (`run`) does **not** have this bug: it spawns a
 `daemon::watcher::FileWatcher` and drives `process_watcher_changes`
-(serve.rs:4231, takes `&SharedIndex`) — snapshot-then-swap, edge-delta graph
+(takes `&SharedIndex`) — snapshot-then-swap, edge-delta graph
 rebuild + warm PageRank (ADR-0165/0166), fresh derived caches — on every
 debounced change batch, atomically swapping a `SharedIndex`. The delta==full
-parity is already tested (serve.rs:6169). So the freshness *primitives* exist and
+parity is already tested. So the freshness *primitives* exist and
 are proven; they are simply not wired into `run_mcp`.
 
 The reuse is **not** verbatim, though: the HTTP watcher loop is an infinite
--- a/docs/adrs/0200-mcp-index-freshness-via-watcher.md
+++ b/docs/adrs/0200-mcp-index-freshness-via-watcher.md
@@ -1,0 +1,140 @@
+---
+id: '0200'
+title: MCP index freshness — wire the file watcher into serve --mcp
+status: ACCEPTED
+date: 2026-07-16
+triggered_by: issue #18 (serve --mcp serves a stale index after file edits)
+loop: planning
+---
+
+# ADR-0200: MCP index freshness — wire the file watcher into `serve --mcp`
+
+## Context
+
+`cxpak serve --mcp` builds its index **once** at startup (`run_mcp` →
+`spawn_mcp_index_build`, `src/commands/serve.rs:2238/2318`, ADR-0185) and never
+refreshes it. Its own doc comment admits it (serve.rs:2448-2449: "a long-lived
+session would keep the first ready index (no periodic rebuild)"). Reproduced
+(REPRO.md): a warm MCP session returns `out_degree 0` for a symbol after an edit
+that a fresh index shows as `out_degree 1` — silently, no staleness signal.
+
+The HTTP path (`run`, serve.rs:1351) does **not** have this bug: it spawns a
+`daemon::watcher::FileWatcher` and drives `process_watcher_changes`
+(serve.rs:4231, takes `&SharedIndex`) — snapshot-then-swap, edge-delta graph
+rebuild + warm PageRank (ADR-0165/0166), fresh derived caches — on every
+debounced change batch, atomically swapping a `SharedIndex`. The delta==full
+parity is already tested (serve.rs:6169). So the freshness *primitives* exist and
+are proven; they are simply not wired into `run_mcp`.
+
+The reuse is **not** verbatim, though: the HTTP watcher loop is an infinite
+`loop {}` (serve.rs:1392-1399) that is detached and only dies on process exit.
+That is fine for a server whose lifetime *is* the process, but the MCP stdio loop
+returns on stdin EOF and is driven in-process by tests — a detached infinite
+thread would leak into every test and cannot be joined on shutdown
+(resource-cleanup). So this decision must also own the watcher's lifecycle.
+
+There is also a **pre-existing second writer** of the readiness cell on the MCP
+path: the opt-in embedding-enrichment phase (`enrich_ready_with_embeddings` →
+`publish_ready_enriched`, serve.rs:2405-2417, ADR-0186) runs on the *same*
+background build thread and, seconds after `Ready`, blind-swaps `ReadyEnriched`
+built from the **startup** index. A watcher that republishes `Ready(edited)` in
+that window would be silently clobbered by the late enrichment (stale, but
+embedding-adorned) — reintroducing exactly the staleness #18 is about. So this
+decision must *coordinate* the watcher with that existing writer, not merely add
+one. (The default no-embeddings path returns early at serve.rs:2375 — the race is
+config-gated and window-bounded, but a correctness defect nonetheless.)
+
+Human decision: it re-opens the ADR-0185 trade-off (that ADR accepted "no
+periodic rebuild" for a short-lived one-task-one-connection model) now that a
+long-lived warm MCP session is a real usage pattern (a persistent commit-gate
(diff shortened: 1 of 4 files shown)


lyubomir-bozhinov:
Adversarial pre-review pass (final) surfaced one Major-but-narrow finding beyond the three acknowledged limitations: `spawn_mcp_watcher` fell back to the uncanonical root on `canonicalize()` failure **silently**. Because `process_watcher_changes` `strip_prefix`es every `notify` event path against that root, an uncanonical root would make *every* event miss — reintroducing #18's silent-staleness mode behind a narrow trigger (repo path becoming un-canonicalizable in the window between the successful build and watcher install). Fail-open direction was already correct; it just wasn't observable.

Fixed in 9d6a81c: log on canonicalize failure (matching the file's existing `FileWatcher::new`-failure log convention), fail-open unchanged. Log-only, no behavior change → no new test. Full pre-commit suite (fmt + clippy + tests) green.

https://claude.ai/code/session_019hJMLmPVoGS2FeUUX4w4Jz

lyubomir-bozhinov:
**Dual-methodology adversarial review round** (`/code-review-pr` + `/code-review-pr-strict`, independent subagents, suite run locally). Verdict: **APPROVE**, no CRITICAL/HIGH. The strict pass hand-traced every {watcher republish, enrich CAS, tools/call snapshot} interleaving and confirmed the CAS is genuinely atomic (single write-lock scope, no TOCTOU) with ABA defeated (base pinned by `ready_base` across the window) — all converge to the fresh index. Four MEDIUM (non-blocking); actioned:

- **Test flake (fixed):** the freshness tests slept a guessed interval then edited once — an edit before the FileWatcher installs was missed with no retry. The edit is now re-asserted inside the bounded poll (all 3 tests), removing the install-timing race; the subprocess `initialize` bound is 5s→30s for cold-start headroom. (The one-off 5s timeout a reviewer saw was confirmed CPU-contention flake, not a defect.)
- **ADR-0200 (fixed):** three stale self-citations that pointed at unrelated code → symbols.
- **Deferred (surfaced for your call, not in this PR):** (1) `republish_watcher_index` does a 2x deep-clone per edit — an `Arc::make_mut` optimization, deferred because it would require re-proving the enrich-CAS ABA invariant for a non-blocking perf win; (2) the pre-existing **HTTP** watcher (`run`) has the same canonicalize gap this PR fixed for MCP — pre-existing and out of #18's MCP scope, recommended as a separate fast-follow.

Local: `cargo test --features daemon,embeddings` full suite 0 failed (subprocess freshness test passes, not ignored); clippy + fmt clean.

https://claude.ai/code/session_019hJMLmPVoGS2FeUUX4w4Jz

PR #29, opened 2026-07-19
fix(watcher): bounded memory for serve --mcp — 3.1.2

lyubomir-bozhinov:
## Why

A long-running `cxpak serve --mcp` grew to **~6 GB RSS with a pegged core** and had to be killed. Root cause is three compounding defects on the watcher rebuild path (see ADR-0204):

1. **The watcher ingested git-ignored files.** The index is built from git-tracked files only, but `classify_changes` had no ignore filter — every path under the watched root (`target/`, `.cxpak/cache/`, `.git/`, `node_modules/`, vendored trees) was fed to `apply_incremental_update` and upserted. Any `cargo build` / git op / cache write grew `index.files` **without bound** and re-triggered full rebuilds forever. (The TypeScript in the original CPU sample was a vendored/generated file that should never have entered the index.)
2. **`build_call_graph` re-parsed each file once per symbol** — `Σ(symbols)` full tree-sitter parses per file, quadratic on large files; catastrophic on the wrongly-ingested giant file.
3. **The full-index deep clone ran before the no-op check** — every spurious wake cloned the whole index, then discarded it.

## What

- `classify_changes` drops git-ignored paths (`git2::is_path_ignored`) + explicit `.git/` guard; fail-open if the repo can't be opened.
- `process_watcher_changes` early-returns **before** the deep clone when no relevant path changed.
- `build_call_graph` parses each file **once** (`extract_calls_by_symbol`), set-equivalent to the old per-symbol path; non-tree-sitter languages keep the regex fallback.
- Version bump 3.1.2 (Cargo.toml/lock, plugin.json, marketplace.json, ensure-cxpak). ADR-0204.

## Validation

- Full suite **3031/3031**, clippy `-D warnings` clean, fmt clean.
- **Whole-repo call graph is byte-identical** between 3.1.1 and 3.1.2 over the 517-file repo (empty diff) — refactor is behavior-preserving.
- New tests: `test_calls_by_symbol_matches_per_symbol_rust`, `test_classify_changes_skips_git_ignored`, `test_process_watcher_changes_ignored_only_skips_rebuild`.
- End-user smoke (overview/trace/onboard) green; release profile compiles.

https://claude.ai/code/session_01KeA5kdfUaF4SgMaqzgutHB

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "cxpak",
       "source": "./plugin",
       "description": "Structured codebase context for Claude. Auto-context on every prompt plus an on-demand single-page dashboard with Cmd+K command palette: three modes — Overview (health, top risks, signals, repo-DNA), Explore (dependency + risk lenses), and History (architecture timeline). 5 intent-parameterized MCP tools, 12 HTTP endpoints, 14 LSP custom methods — all wired to real intelligence: health scoring, blast radius, change prediction, drift detection, security surface, data flow tracing, cross-language resolution. tree-sitter across 43 languages. Cross-process deterministic output.",
-      "version": "3.1.1"
+      "version": "3.1.2"
     }
   ]
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -982,7 +982,7 @@ dependencies = [
 
 [[package]]
 name = "cxpak"
-version = "3.1.1"
+version = "3.1.2"
 dependencies = [
  "assert_cmd",
  "axum",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cxpak"
-version = "3.1.1"
+version = "3.1.2"
 edition = "2021"
 rust-version = "1.91"
 description = "Spends CPU cycles so you don't spend tokens. The LLM gets a briefing packet instead of a flashlight in a dark room."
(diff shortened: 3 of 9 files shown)


lyubomir-bozhinov:
### Adversarial review round (2 independent reviewers) + fix

Ran two adversarial reviewers over the diff before requesting merge.

**Call-graph single-parse refactor — cleared.** 16 adversarial fuzz probes across all 9 tree-sitter languages (nested same-name fns, sibling impl methods, overloads, arrow IIFEs, C# local-function shadowing, Go receiver methods, …) + a structural containment proof; no divergence from the old per-symbol path. Corroborated by a whole-repo call-graph diff (3.1.1 vs 3.1.2) that is **byte-identical**.

**One real live bug found and fixed** (commit `d83b44a`): `classify_changes` filtered only git's ignore rules, but `Scanner::scan` also excludes `BUILTIN_IGNORES` + `.cxpakignore`. Committed build-noise that git does *not* ignore — `Cargo.lock`, `*.png`, `dist/`, `*.min.js`, `.DS_Store` — would have leaked into the live index and re-triggered a rebuild on every `cargo build` (same failure class as the original bug, different file set). Now reuses the scanner's exact exclusion set; regression test `test_classify_changes_skips_builtin_noise_not_gitignored`. Also hardened `Repository::open`→`discover` (defensive; `Scanner::new` already guarantees the watch root is a repo root).

**Validation:** full suite **3133 pass / 0 fail**, clippy `-D warnings` clean, fmt clean, tarpaulin **93.45%** (>90% gate). End-user smoke (overview/trace/onboard) green; release profile compiles.

PR #48, opened 2026-07-23
fix(watcher): one deep clone per edit batch, Arc-share embedding matrix

### LINKED ISSUE #47: cxpak serve --mcp: unbounded memory growth (57GB) from two full CodebaseIndex deep clones per watcher batch
**Severity: High** (memory — `cxpak serve --mcp` grows unbounded)

## What happened

A long-lived `cxpak serve --mcp .` process (spawned by Claude Code) grew to a **57 GB physical footprint (peak 64 GB)** after ~44 min of a normal editing session, and was still climbing ~1 GB/min. It had exhausted system swap (61.5 GB of 62.5 GB used). RSS reads low (~1.8 GB) only because macOS had paged ~57 GB of dirty heap out to swap.

`vmmap -summary <pid>`:

```
Physical footprint:         57.2G   (peak 64.1G)
Writable regions: Total=64.4G written=57.2G(89%) resident=1.0G(2%) swapped_out=56.2G(87%)

REGION TYPE            VIRTUAL   RESIDENT   DIRTY   SWAPPED   COUNT
MALLOC_LARGE            16.1G     405.5M   405.5M    14.3G     142
MALLOC_SMALL           44.3G       1.2G     1.2G    42.5G   11331
```

Two `vmmap` snapshots ~1 min apart: footprint 57.2G → 59G, `MALLOC_SMALL` 43.3G → 44.3G (11087 → 11331 regions), all dirty. CPU was low (~3%), 5 threads mostly idle in `mach_msg` — this is not a busy loop, it grows per file-change event. macOS munmaps freed `MALLOC_LARGE` blocks, so 16 GB retained there indicates large buffers held **live**, not merely churned.

## Root cause (mapped)

The MCP freshness-watcher path re-materializes the entire `CodebaseIndex` — **twice** — on every debounced watcher batch, and reconstructs the ignore matcher each time. For an active session (constant real source edits), this fires roughly once per second and sustains heavy large-allocation churn/retention.

Per batch, in `spawn_mcp_watcher`'s loop (`src/commands/serve.rs:2635-2648`):

1. **First full deep clone** — `process_watcher_changes` (`src/commands/serve.rs:4528`):
   ```rust
   let mut next: CodebaseIndex = (*snapshot).clone();
   ```
   Clones the *entire* index including file contents, parse results, graph, call graph — and the **embeddings matrix** (the `CodebaseIndex` clone at 4528 carries `embedding_index`; the large `MALLOC_LARGE` buffers).

2. Then `rebuild_graph_delta` + `compute_pagerank_seeded` (100 iters) + full `build_test_map` + `update_conventions_incremental` on that clone.

3. **Second full deep clone** — `republish_watcher_index` (`src/commands/serve.rs:2540-2543`):
   ```rust
   let mut idx = match shared.read() {
       Ok(g) => (**g).clone(),        // second full CodebaseIndex deep clone
       ...
   ```
   (It only nulls `embedding_index` *after* the base clone in `process_watcher_changes` already cloned it.)

4. **Ignore matcher rebuilt from disk every batch** — `classify_changes` (`src/commands/watch.rs:103-114`) calls `git2::Repository::discover` and rebuilds a fresh `GitignoreBuilder` (all `BUILTIN_IGNORES` + re-reads `.cxpakignore`) on every invocation.

The ignored-only early-return (`serve.rs:4515`) correctly prevents a `.cxpak/`-cache self-trigger loop, so growth requires real edits — but under an actual coding session those are continuous, and two whole-index clones (incl. embeddings) per edit is enough to drive footprint into the tens of GB.

## Steps to reproduce

1. `cxpak serve --mcp .` on a moderately large repo (ideally with an `.cxpak.json` `embeddings` section so the embedding matrix is present).
2. Drive continuous source-file edits under the watched root for ~30–45 min (e.g. an active Claude Code / editor session).
3. `vmmap -summary <pid> | grep -i footprint` — observe physical footprint climbing into the tens of GB and not receding.

## Suggested fixes

- Eliminate the redundant second deep clone: `process_watcher_changes` already produced the new `Arc<CodebaseIndex>` it swapped into `shared`; `republish_watcher_index` should reuse/`Arc::clone` that instead of `(**g).clone()`-ing the whole index again.
- Avoid cloning the embeddings matrix on the content-edit path — the republished watcher index drops `embedding_index` anyway; strip/skip it *before* the first clone rather than deep-copying it every batch.
- Prefer in-place mutation of the swapped index over `(*snapshot).clone()` per batch, or reuse a single working buffer.
- Cache the `git2::Repository` handle + gitignore matcher across batches in `classify_changes` instead of rebuilding from disk each time.
- Investigate whether old `Arc<CodebaseIndex>` snapshots are being retained (slow/leaked tool-call handlers holding `snapshot_ready_index` clones) — the retained 16 GB of `MALLOC_LARGE` suggests live retention beyond pure churn.

## Environment

- cxpak 3.1.2 (Homebrew), commit `5d0066a`
- macOS 26.5 (25F71) arm64
- Observed pid footprint peak 64.1 GB

## Related

Overlaps with #44 (watcher/daemon lifecycle: unbounded channel, stale rename entries, LSP watcher leak) but is a distinct code path — the sustained growth here is the MCP `serve --mcp` per-batch double deep clone, not the channel backlog burst.

cefothe:
## What & why

`cxpak serve --mcp` grew to a **57 GB physical footprint (peak 64 GB)** during a normal editing session, climbing ~1 GB/min. The MCP freshness watcher deep-cloned the entire `CodebaseIndex` **twice per real-edit batch**, and each clone deep-copied all `IndexedFile.content` **and** the embedding `Vec<f32>` matrix by value. Field report: 16 GB in `MALLOC_LARGE` (embedding matrices) + 44 GB in `MALLOC_SMALL` (file contents).

This PR lands the full fix (P0 + P1 + P2 from the analysis):

### P0 — one clone per batch + Arc-shared embedding matrix
- `embedding_index: Option<EmbeddingIndex>` → `Option<Arc<EmbeddingIndex>>`: cloning a `CodebaseIndex` bumps a refcount instead of copying `Vec<f32>` — **eliminates the ~16 GB `MALLOC_LARGE` class**.
- `process_watcher_changes` clears `embedding_index` once before the swap and returns the swapped `Arc`; `spawn_mcp_watcher` republishes that same `Arc`. `republish_watcher_index` (the redundant second whole-index clone) is **deleted**.

### P2 — copy-on-write file storage
- `CodebaseIndex.files: Vec<IndexedFile>` → `Vec<Arc<IndexedFile>>`. Cloning the index shares unchanged files as refcount bumps, so the surviving per-batch clone copies **O(delta)** file contents, not O(codebase) — **collapses the 44 GB `MALLOC_SMALL` growth**.
- CoW by construction: `upsert_file` = `remove_file` (`swap_remove` the `Arc`) + `push(Arc::new(..))`; `remove_file` drops an `Arc`. No `IndexedFile` is mutated in place, so an old snapshot a reader still holds is never corrupted. Nine `&[IndexedFile]` helper signatures became `&[Arc<IndexedFile>]` (bodies unchanged, read sites auto-deref).

### P1 — memory-ceiling proof
- `watcher_does_not_accumulate_index_versions_across_batches` drives 25 real-edit batches and asserts `Arc::strong_count` on the shared/readiness Arc returns to a fixed baseline every batch — **demonstrates** (not just asserts) the bounded footprint.

Preserves the ADR-0200 6-signal-fallback and snapshot-then-swap reader-consistency invariants. Design + rejected alternatives in **ADR-0205**; predecessor is ADR-0204.

Closes #47

## How it was verified

Toolchain: rustc **1.91.1** (project MSRV 1.91; system default was 1.87, installed 1.91).

- `cargo build` — green.
- `cargo fmt -- --check` — clean.
- `cargo clippy --lib -- -D warnings` — clean on all changed code.
- **`cargo test` — full suite green:** lib **2587 passed / 0 failed / 3 ignored**, and every integration binary passes. (One pre-existing, unrelated flake — `install_then_real_git_merge_union_resolves_artifact` in `hook_integration.rs` — fails identically at the base commit; it does real git checkouts in a temp repo and touches none of this code.)

New tests (each fails on the pre-fix code by construction):
- `process_watcher_changes_clears_embedding_and_publishes_single_arc` — `Arc::ptr_eq` proves a single publish, not two clones.
- `embedding_index_clone_shares_matrix_not_copies` — `Arc::strong_count`; **cannot compile against the pre-fix non-`Arc` field**.
- `watcher_does_not_accumulate_index_versions_across_batches` — the P1 ceiling proof over 25 batches.
- `process_watcher_changes_ignored_only_publishes_none` — spurious wakes allocate no version.

## Checklist

- [x] `cargo fmt -- --check` is clean
- [x] `cargo clippy` clean on all changed code (`--lib -D warnings`); see note re `--all-targets`
- [x] `cargo test` passes (2587 lib + all integration binaries); new code covered with 4 new tests
- [x] Feature-gated change (`embeddings`): default build + tests green
- [x] Docs updated — **ADR-0205** added under `docs/adrs/` and indexed
- [x] Architecturally significant → ADR added
- [x] Commits follow Conventional Commits, logically scoped (P0 fix; P1/P2 perf follow-up)

## Notes for reviewers

- **Memory outcome:** P0 eliminates the 16 GB embedding class; P2 turns the residual per-batch file-content clone from O(codebase) into O(delta). Together this bounds both `MALLOC_LARGE` and `MALLOC_SMALL`; the P1 test demonstrates the ceiling. Output is byte-identical — the ADR-0204 delta-parity guard still passes.
- **Pre-existing clippy lint (not from this PR):** `cargo clippy --all-targets -- -D warnings` fails on `tests/spa_render.rs:363` (`search_is_some`) — identical code on `main`, surfaced only by clippy 1.91.1. Out of scope; happy to fix separately.
- **CoW correctness:** the delta path never mutates a shared `IndexedFile` in place (all `Arc::make_mut` uses are in `#[cfg(test)]` fixtures with refcount 1). Old snapshots held by in-flight readers stay valid.
- **Remaining follow-up (ADR-0205 "Revisit if"):** an audit of `snapshot_ready_index` handler retention if a concurrent MCP handler model is ever added; and hoisting the `classify_changes` git2/gitignore matcher (CPU-only, Fix #4).

### DIFF
--- a/docs/adrs/0205-watcher-single-clone-arc-embeddings.md
+++ b/docs/adrs/0205-watcher-single-clone-arc-embeddings.md
@@ -1,0 +1,131 @@
+---
+id: '0205'
+title: Watcher single-clone — one deep clone per edit batch, Arc-shared embedding matrix
+status: ACCEPTED
+date: 2026-07-23
+triggered_by: field report (issue #47) — `cxpak serve --mcp` grew to a 57 GB physical footprint (peak 64 GB) after ~44 min of a normal editing session, climbing ~1 GB/min
+loop: implementation
+---
+
+# ADR-0205: Watcher single-clone — one deep clone per edit batch, Arc-shared embedding matrix
+
+## Context
+
+ADR-0204 bounded the watcher's *ingestion set* and guarded the deep clone on
+**spurious** wakes (all-ignored batches early-return before the clone). It did
+**not** touch the **real-edit** path, where each debounced batch that carries a
+genuine source change still deep-cloned the entire `CodebaseIndex` **twice**:
+
+1. `process_watcher_changes` (`serve.rs`) — `let mut next = (*snapshot).clone()`,
+   the working copy the delta rebuild mutates and swaps into the `shared` mirror.
+2. `republish_watcher_index` (`serve.rs`) — `(**g).clone()` of that just-swapped
+   index **a second time**, solely to produce a copy with `embedding_index = None`
+   (the ADR-0200 6-signal fallback) to publish into the `readiness` cell.
+
+`CodebaseIndex` derives `Clone` with no structural sharing (`core_graph/index.rs`),
+so each clone is a full deep copy of every `IndexedFile.content` String,
+`term_frequencies`, `graph`, `call_graph`, **and** the embedding matrix
+(`EmbeddingIndex.vectors: Vec<f32>`). Under a continuous editing session (roughly
+one batch/second) this sustained two whole-index deep copies per edit — with the
+embedding matrix copied on both — driving the physical footprint into the tens of
+GB. A `vmmap` field report showed 16 GB retained in `MALLOC_LARGE` (the flat
+`Vec<f32>` matrices) and 44 GB in `MALLOC_SMALL` (file contents), dirty and
+swapped — live retention plus allocator churn, not a bounded working set.
+
+Two mechanisms compounded the cost:
+- The **redundant second whole-index deep clone** existed only to null one field.
+- The **embedding matrix was deep-copied by value** on every `CodebaseIndex`
+  clone, because `embedding_index` was `Option<EmbeddingIndex>` (owned by value).
+
+## Options considered
+
+- **Option A — strip embeddings before the swap + Arc-share the matrix (chosen):**
+  clear `embedding_index` once, inside `process_watcher_changes`, immediately
+  before `Arc::new(next)`, and return that freshly-swapped `Arc` so the watcher
+  publishes the *same* pointer into `readiness` — deleting the second clone
+  entirely. Independently, change the field to `Option<Arc<EmbeddingIndex>>` so
+  cloning a `CodebaseIndex` bumps a refcount for the matrix instead of copying
+  `Vec<f32>`. The two changes ship together: stripping-before-swap alone keeps
+  the ADR-0200 invariant, and the `Arc` wrap removes the matrix from the *first*
+  (unavoidable) clone's cost as well as the enrichment clone's.
+- **Option B — make `republish_watcher_index`'s clone cheap via the `Arc` field:**
+  keep the second clone but rely on the `Arc`-wrapped matrix to make it O(1).
+  Rejected: Arc-wrapping *only* the embedding field does not make the whole-index
+  `(**g).clone()` cheap — `files`/`content`, `term_frequencies`, and `graph` are
+  still deep-copied. It would leave the 44 GB `MALLOC_SMALL` churn fully in place.
+- **Option C — in-place mutation of the swapped index (issue #47 suggestion 3):**
+  mutate the `Arc`'s pointee instead of cloning. Rejected: violates the
+  snapshot-then-swap reader-consistency invariant (`serve.rs`, the comment block
+  above the clone) — long-running MCP tool handlers alias the pre-swap `Arc`, so
+  in-place mutation is a torn-read hazard for no benefit over Option A.
+
+## Decision
+
+Adopt Option A.
+
+- `embedding_index: Option<Arc<crate::embeddings::EmbeddingIndex>>`
+  (`core_graph/index.rs`). Read sites auto-deref through `Arc`
+  (`relevance/signals.rs`); `has_embedding_index()` is unchanged; the two
+  by-value `Some(emb)` writes become `Some(Arc::new(emb))`
+  (`serve.rs` `publish_ready_enriched`, plus test constructors). `CodebaseIndex`
+  derives only `Debug, Clone` (no `serde`), so the wrap introduces no
+  serialization change; `EmbeddingIndex` keeps its own `Serialize/Deserialize`
+  and `load`/`save`, which never run through the index.
+- `process_watcher_changes` now clears `embedding_index` once before the swap
+  and returns `Option<Arc<CodebaseIndex>>` — `Some(new_arc)` on a real change,
+  `None` on the ignored-only / no-op / poisoned paths. `spawn_mcp_watcher`
+  publishes that exact `Arc` into `readiness`. `republish_watcher_index` is
+  **deleted** (its clearing responsibility moved into the pre-swap step).
+
+Covered by `process_watcher_changes_clears_embedding_and_publishes_single_arc`
+(published index is embeddings-free **and** `Arc::ptr_eq` with the `shared`
+mirror — proving a single clone), `embedding_index_clone_shares_matrix_not_copies`
+(`Arc::strong_count` proves the matrix is shared, not copied — and the test
+cannot compile against the pre-fix non-`Arc` field), and
+`process_watcher_changes_ignored_only_publishes_none`. The ADR-0204 parity guards
+(`process_watcher_changes_delta_parity_with_full_rebuild`, `classify_changes`
+tests) remain green.
+
+## Consequences
+
+### Positive
+- **One** whole-index deep clone per real-edit batch instead of two — halves the
+  `MALLOC_SMALL` (file-content) churn on the hot path.
+- The embedding matrix is never deep-copied on the watcher path: per-batch matrix
+  copies drop from three (the two watcher clones plus the enrichment clone) to
+  zero — the ~16 GB `MALLOC_LARGE` class is eliminated.
+- The ADR-0200 6-signal-fallback invariant is preserved and now enforced at a
+  single, testable point (pre-swap) rather than via a redundant clone.
+
+### Negative
+- The embeddings-clearing responsibility is no longer named by a dedicated
+  `republish_watcher_index` function; it is a documented step inside
+  `process_watcher_changes`. Auditing the invariant now means reading that step
+  (called out in-line with an ADR-0200/0205 comment) rather than a standalone fn.
+- `process_watcher_changes` gained a return value; the HTTP watcher
+  (`serve.rs`) and LSP backend (`lsp/backend.rs`) call it as a statement and
+  discard the `Option` — intentional, as they own no `readiness` cell.
+
+### Neutral
+- The remaining per-batch clone still deep-copies all `IndexedFile.content`
+  (the residual `MALLOC_SMALL` class). Delta rebuilds touch only a handful of
+  files, so this copy is mostly redundant, but bounding it is a separate change.
+- The `Arc::strong_count`-returns-to-baseline assertion demonstrates no *extra*
+  index version survives a batch; it does not, by itself, prove no *external*
+  holder pins an old version (see Revisit if).
+
+## Revisit if
+- **`MALLOC_SMALL` still dominates under load.** The residual full-content clone
+  per batch is the next ceiling. Move to `files: Vec<Arc<IndexedFile>>` (or an
+  immutable/`im::Vector`) copy-on-write so a delta clone shares unchanged files
+  and copies only the changed ones — turning the per-batch cost from
+  O(codebase) to O(delta). Tracked as a separate ticket (issue #47 P2).
+- **An old `Arc<CodebaseIndex>` is pinned by a slow/hung MCP tool handler**
+  (`snapshot_ready_index` hands each call an `Arc::clone`). If a handler stalls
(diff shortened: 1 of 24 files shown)


cefothe:
## PR Review — one deep clone per edit batch, Arc-share embedding matrix

**Verdict: ✅ APPROVE** · Composite **9.2/10** · Phase 0 security scan **CLEAN**

Reviewed across six dimensions (Rust correctness, performance, architecture, tests, security, cross-file ripple). All findings verified against source, not just the diff.

### Security-critical scan — CLEAN (not blocked)
No hardcoded secrets (`token_count` is a struct field), zero URL changes, no `unsafe`/`process`/`exec`/`transmute`, no `Cargo.toml`/`Cargo.lock` dependency changes.

### Specialist scores
| Dimension | Score | Verdict |
|---|---|---|
| Rust correctness | 9/10 | clean |
| Performance | 9/10 | claims real, not just asserted |
| Architecture | 9/10 | invariants hold |
| Tests | 2.6/3.0 | meaningful, not cheater tests |
| Security | 9.5/10 | PASS |
| Cross-file ripple | 9/10 | mechanical |

### What this PR gets right
- **Root cause eliminated.** The two per-batch whole-index deep clones (`process_watcher_changes` + the deleted `republish_watcher_index`) collapse to one, and the survivor is copy-on-write (`Vec<Arc<IndexedFile>>`) — file-content copy drops from O(codebase) → O(delta). The embedding `Vec<f32>` matrix (`Option<Arc<EmbeddingIndex>>`) is refcount-shared, never deep-copied on the watcher path. Directly targets both the 44 GB `MALLOC_SMALL` and 16 GB `MALLOC_LARGE` classes from the field report.
- **COW safety proven, not asserted.** `upsert_file`/`remove_file` are `swap_remove`+`push(Arc::new(..))` — never in-place mutation — so a slow MCP reader holding an old snapshot is never corrupted. No `Arc::get_mut` anywhere; all 5 `Arc::make_mut` sites are test-only.
- **ADR-0200 invariant holds on every path.** The embedding-clear moved inline but sits before the sole `shared.write()`; all three early `return None` exits precede any swap.
- **Genuinely strong tests.** `embedding_index_clone_shares_matrix_not_copies` is a compile-time guard (won't build against the pre-fix field) plus an `Arc::strong_count` check; the memory-ceiling test would fail pre-fix (baseline count 2 not 3, no `ptr_eq`).
- **ADR-0205 is exemplary** — options-considered with rejection rationale, and honest Consequences/Revisit-if disclosing the unmitigated (bounded) slow-handler pin.

### Non-blocking findings (all NIT / WARNING)

1. **[NIT — flagged independently by 3 reviewers] Poisoned-write-lock divergence** · `src/commands/serve.rs:4569-4577`. On a poisoned `shared.write()`, `if let Ok(mut g)` silently skips the swap but the fn still returns `Some(new_arc)` — the MCP caller then publishes into `readiness` an `Arc` that is *not* in `shared`, diverging the mirror. The read-lock path (`:4510`) returns `None` for symmetry; the write path doesn't. Unreachable in practice (requires a prior mid-write panic by the sole writer). Suggested fix:
   ```rust
   match shared.write() {
       Ok(mut g) => *g = Arc::clone(&new_arc),
       Err(_) => return None,
   }
   ```

2. **[WARNING] No `#[must_use]` on `process_watcher_changes`** · `src/commands/serve.rs:4489`. Two callers (HTTP `run`, LSP backend) correctly discard the `Option` as a bare statement, but the "safe to discard only if you own no readiness cell" rule lives only in the ADR. A future readiness-owning caller could silently drop the republish with no compiler signal. Add `#[must_use = "callers owning a readiness cell must publish the returned Arc"]` and make the two discards explicit `let _ = …`.

3. **[WARNING — follow-up, not this PR] `term_frequencies` per-batch copy** · `src/core_graph/index.rs:33`. Still deep-copied each batch; on >50k-file token-dense repos it becomes the dominant per-batch *transient* (bounded churn, freed each batch — cannot regrow the leak). The ADR correctly names it the next `Arc`/delta candidate; worth tracking on a large-repo profile.

4. **[NIT] Coverage gaps** — the poisoned-lock branch and `update_count == 0` early-return are untested; the memory-ceiling proof drives single-file batches only (multi-file delta covered indirectly by `tests/parity.rs`).

5. **[NIT] Deref-adapter noise** — `.map(|f| has_inline_tests(f))` closures (`risk.rs`, `render.rs`) and `Vec<&Arc<IndexedFile>>` annotations (`health.rs:156`, `api_surface.rs:108`) are correct but a future "simplify" could break the build; a `// Arc deref` comment would guard them.

### CLAUDE.md compliance
No TODO/FIXME, no placeholders/stubs, no dead code (the deleted fn is fully removed), tests shipped alongside, resource cleanup preserved, error paths handled. Atomic and focused on issue #47.

**Recommendation:** merge as-is, or fold in findings #1 and #2 first — both are ~2-line hardening changes that convert two convention-enforced seams into compiler-enforced ones.

<sub>Automated multi-agent review (`/code-review-pr`). Findings verified against source at `678a512`.</sub>

lyubomir-bozhinov (dismissed):
## Strict review — APPROVE (0 blocking)

Reviewed adversarially across three independent lenses (CoW/concurrency correctness, memory/refcount-leak, blast-radius/parity/test-adequacy), each verifying against the real code — build, `clippy --lib --tests -D warnings`, the full 2587-test lib suite, and the 4 new tests all green; fail-first confirmed by grafting the new tests onto a `main` worktree (real compile errors pre-fix).

### Verified (empirically, not on faith)
- **P0 single-publish** — one deep clone per batch; `process_watcher_changes` returns the swapped `Arc` and the MCP loop republishes *that same* Arc into `readiness` (`Arc::ptr_eq` asserted across 25 batches). `republish_watcher_index` (the redundant second whole-index clone) is gone. HTTP/LSP callers correctly discard the return (they publish via `shared` internally).
- **P2 CoW is real** — built a 20-file on-disk index, edited 3: **17/17 untouched files stay pointer-shared, 3/3 changed files freshly allocated** → `MALLOC_SMALL` bounded to O(delta). `embedding_index: Arc` eliminates the ~16 GB `MALLOC_LARGE` class.
- **No in-place mutation of shared data** — `upsert_file` = `remove_file` + `push(Arc::new(…))`; `remove_file` drops via `swap_remove`; zero `Arc::get_mut`; all 5 `Arc::make_mut` are test-only on refcount-1 fixtures (brace-depth scan). `IndexedFile` has no interior mutability; no `unsafe` introduced.
- **Arc-identity hazard impossible** — `IndexedFile` derives only `Debug, Clone` (no `Eq`/`Hash`/`Ord`), so no value-vs-pointer comparison exists to break.
- **Reader consistency** — snapshot-then-swap; a reader's cloned `Arc` snapshot is frozen, unaffected by a concurrent batch swap.
- **No leak today** — every Arc holder drops its prior generation each batch; the MCP stdio loop is sequential. Builds with `embeddings` on **and** off.

### Non-blocking (please consider; none block merge)
1. **P1 test comment overclaims (top note).** `Arc::strong_count == 3` proves *no divergent double-clone / current-generation holder count* — not total-heap boundedness (a hypothetical version-history `Vec<Arc>` would still pass it). Suggest softening the "demonstration … bounded" wording, or pairing with an allocation-count assertion.
2. **"Fails on pre-fix" is a compile failure** for 3 of 4 tests (the strongest possible guard for a type change) — the changelog phrasing implying runtime-divergence detection is slightly loose.
3. **Minor coverage gap** — no single test drives a batch *mixing* modifies + deletes through `process_watcher_changes`, and the ceiling test runs without embeddings present. Cosmetic (the parity proptest covers mixed add/remove; the ceiling is orthogonal to embeddings).
4. Pre-existing & correctly disclosed: the `--all-targets` clippy lint on `tests/spa_render.rs:363` and the `hook_integration.rs` flake are on untouched files; the `snapshot_ready_index` retention follow-up is a correctly-scoped hypothetical (only under a future concurrent MCP handler model), not a present leak.

Everything load-bearing is correct and empirically verified. Mergeable as-is; the only thing I'd tweak is the P1 test's doc comment (non-blocking).

lyubomir-bozhinov (approved):

Issue #19, opened 2026-07-16
MCP tools/call: no machine-readable discriminator between retryable 'indexing in progress' and terminal errors


lyubomir-bozhinov:
## Summary

On the **MCP** surface, `tools/call` returns `isError: true` with a human‑readable `text` for at least
three semantically different conditions, with **no machine‑readable discriminator** between them:

1. **Retryable** — the background index is still building (`"cxpak: indexing in progress — … Retry this
   call in a few seconds."`).
2. **Terminal** — the index build failed (`"cxpak: indexing failed … restart the MCP server to retry."`).
3. **Terminal** — parameter‑validation / capability errors.

A client can only tell "retry in a moment" from "give up / restart" by **substring‑matching the prose**.

## Evidence (source, `src/commands/serve.rs`, 3.1.0)

- `snapshot_ready_index` maps readiness to a `Result<_, String>`:
  - `IndexReadiness::Building => Err(INDEXING_IN_PROGRESS_MESSAGE)` — `serve.rs:2436`
  - `IndexReadiness::Failed(msg) => Err("cxpak: indexing failed … restart …")` — `serve.rs:2437`
- Every `Err` is wrapped by `mcp_tool_error` (`serve.rs:4213`) into
  `{"result":{"content":[{"type":"text","text": …}],"isError":true}}` — the **same** shape used for
  parameter/capability errors. So `isError:true` alone does not distinguish retryable from terminal.

## Impact

Consumers must hardcode the exact human strings. Concretely, our client greps for
`"indexing in progress"` and pins that exact substring in a unit test **specifically because a wording
change would silently disable retry** — and it still can't distinguish `Failed` (don't retry; restart)
from `Building` (retry) unless it also hardcodes the failure string. A client that hardcodes neither
either treats a still‑warming server as a hard error, or retries a genuinely failed build forever.

## Reproduction

1. `cxpak serve --mcp <a repo large enough that indexing takes a moment>`.
2. From an MCP client: `initialize`, then a `tools/call` immediately.
3. Response is `isError:true` with `text` = "cxpak: indexing in progress …". Nothing in the envelope
   marks it as *retryable* vs the `Failed`/param‑error cases, which are also `isError:true`.

## Suggested fix

Attach a machine‑readable discriminator to the not‑ready / error results — e.g. MCP
`CallToolResult.structuredContent` (or `_meta`) carrying `{"status":"indexing","retryable":true}` /
[...]

PR #50, opened 2026-07-27
perf(watcher): jemalloc allocator + bounded/debounced watcher to cap RSS

cefothe:
## Problem

The MCP server (`cxpak serve --mcp`) climbed past **35 GB RSS** while watching a large multi-module Maven repo during review-time file churn. This was diagnosed as a recurrence of the earlier watcher memory issue (#47 / ADR-0205).

**There is no classic reference leak** — the index working set is bounded and `Arc::strong_count` returns to baseline every batch (the P1 ceiling test still passes). But two independent mechanisms let RSS ratchet up without bound:

### Mechanism A — allocator retention (why RSS *stays* at 35 GB)
The binary used the default system allocator (macOS `libmalloc` / glibc), which parks freed blocks in per-process free lists under the watcher's per-batch clone churn. Freed pages are never returned to the OS, so RSS ratchets up and stays there even though the live set is <1 GB.

### Mechanism B — unbounded event backlog (transient, real growth)
`FileWatcher` used a raw **unbounded** `mpsc` channel and watched the whole tree recursively. A git checkout / build touching tens of thousands of files (every `target/`, `.git/`, `node_modules/` path) queued GBs of `PathBuf`s faster than the single watcher thread could drain, and each relevant batch re-triggered a full multi-second reindex.

## Fix

**Mechanism A** — `tikv-jemallocator` as `#[global_allocator]` with aggressive decay (`dirty_decay_ms:1000, muzzy_decay_ms:0, background_thread:true`), cfg-gated off MSVC. jemalloc purges freed pages back to the OS, capping steady-state RSS near the live set. (On macOS `background_thread` is ignored but decay still purges synchronously on free; the background thread benefits the Linux/Docker prod path.)

**Mechanism B + watch-layer filtering** — in `FileWatcher`:
- **Bounded** `sync_channel(65_536)` with non-blocking `try_send` on the FS thread. On overflow an atomic flag drives a `build_index` **full-rebuild fallback**, so no edit is silently lost.
- Coarse `is_noise_path` **pre-filter at the notify callback** drops known noise trees (`.git`, `target`, `node_modules`, `.cxpak`, `dist`, `build`, `.gradle`, `.idea`, `.next`, `.venv`) before they enter the channel. `classify_changes` remains the authoritative correctness filter.
- `collect_debounced()` replaces the ad-hoc `recv+sleep+drain`: it windows (400 ms quiet, 2 s cap) and coalesces duplicate `(path, kind)` events, so a 40k-file checkout becomes **one** rebuild.

Both the HTTP (`run`) and MCP (`spawn_mcp_watcher`) watcher loops use the new path. `notify-debouncer-full` was intentionally **not** used — the in-house bounded+coalesced approach adds no dependency and keeps the tested `classify_changes` pipeline intact.

## Design notes
- Overflow is correctness-safe: dropped non-noise events force a full `build_index` rescan rather than trusting a lossy delta.
- The coarse pre-filter only drops what `classify_changes` already rejects — verified by unit tests.

## Testing
- Full library suite: **2592 passed / 0 failed / 3 ignored** (rust 1.91).
- Includes the P1 ceiling test (`watcher_does_not_accumulate_index_versions_across_batches`), all `process_watcher_changes_*`, and all `mcp_watcher_*` tests.
- New unit tests: noise-filter coverage, `(path,kind)` coalescing, bounded-channel overflow flag.
- `scripts/measure_rss.sh` added for per-second RSS sampling (+ `vmmap`/`leaks` pointers) to confirm before/after.

## Not in this PR (follow-up)
- Finish CoW on `term_frequencies` / `graph` / `call_graph` / `conventions` and remove the redundant per-batch conventions recompute — shrinks the per-batch churn itself. Lower priority now that the allocator caps retention.

## Verify empirically
Rebuild release, restart the MCP server, run `scripts/measure_rss.sh <pid>` during a review, and confirm RSS plateaus in low single-digit GB instead of climbing to 35.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1015,6 +1015,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "tiktoken-rs",
+ "tikv-jemallocator",
  "tiny_http",
  "tokenizers",
  "tokio",
@@ -4190,6 +4191,26 @@ dependencies = [
  "parking_lot",
  "regex",
  "rustc-hash 1.1.0",
+]
+
+[[package]]
+name = "tikv-jemalloc-sys"
+version = "0.6.1+5.3.0-1-ge13ca993e8ccb9ba9847cc330696e02839f328f7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd8aa5b2ab86a2cefa406d889139c162cbb230092f7d1d7cbc1716405d852a3b"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
+name = "tikv-jemallocator"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0359b4327f954e0567e69fb191cf1436617748813819c94b8cd4a431422d053a"
+dependencies = [
+ "libc",
+ "tikv-jemalloc-sys",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,16 @@ tokio-postgres-rustls = { version = "0.13", optional = true }
 rustls = { version = "0.23", optional = true, default-features = false, features = ["ring"] }
 mysql_async = { version = "0.37", optional = true, default-features = false, features = ["minimal", "rustls-tls"] }
 
+# Global allocator (perf/watcher-allocator-debounce, Mechanism A). The default
+# system allocator (macOS libmalloc / glibc) parks freed blocks in per-process
+# free lists under the watcher's per-batch clone churn, so RSS ratchets to tens
+# of GB and never returns. jemalloc with aggressive decay (`dirty_decay_ms` /
+# `muzzy_decay_ms`, see `malloc_conf` in `src/main.rs`) purges those pages back
+# to the OS, capping steady-state RSS near the live set. jemalloc does not build
+# on MSVC, so the dep and the `#[global_allocator]` are both `cfg`-gated off it.
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
+tikv-jemallocator = { version = "0.6", features = ["unprefixed_malloc_on_supported_platforms", "background_threads"] }
+
 [dev-dependencies]
 # Self dev-dependency so integration tests (tests/*.rs) see `cxpak::test_support`
 # with the `test-support` feature on, while the release build never enables it.
--- a/scripts/measure_rss.sh
+++ b/scripts/measure_rss.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# Prints a timestamped RSS (MB) sample once per second until the process exits.
+# Dirty-vs-live split: vmmap --summary <pid>  or  MALLOC_NANOZONE=0 leaks <pid>
+set -euo pipefail
+
+if [[ $# -lt 1 ]]; then
+    echo "Usage: $0 <pid>" >&2
+    exit 1
+fi
+
+PID="$1"
+
+while kill -0 "$PID" 2>/dev/null; do
+    RSS_KB=$(ps -o rss= -p "$PID" 2>/dev/null || true)
+    if [[ -z "$RSS_KB" ]]; then
+        break
+    fi
+    RSS_MB=$(( RSS_KB / 1024 ))
+    printf '%s  pid=%s  rss=%s MB\n' "$(date -u +"%Y-%m-%dT%H:%M:%SZ")" "$PID" "$RSS_MB"
+    sleep 1
+done
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -2653,6 +2653,30 @@ pub fn spawn_mcp_watcher(
                     "cxpak: MCP watcher event buffer overflowed — \
                      delta may be stale; triggering full rebuild"
                 );
+                // Release BOTH strong references to the current index *before*
+                // building its replacement. Building first and swapping after
+                // keeps two full indexes live at once, and this path fires
+                // precisely during an event storm — the moment the process can
+                // least afford to double its largest allocation.
+                //
+                // The trade is availability for memory, and it is the right way
+                // round here: overflow means events were dropped, so we have
+                // just declared the current index lossy. Answering `Building`
+                // ("indexing in progress, retry") is an honest answer to a
+                // question we can no longer answer correctly, it is the same
+                // state the initial background build publishes (ADR-0185), and
+                // MCP clients already retry on it (#19). Serving data we have
+                // just labelled untrustworthy would be the worse choice.
+                match readiness.write() {
+                    Ok(mut g) => *g = IndexReadiness::Building,
+                    Err(p) => *p.into_inner() = IndexReadiness::Building,
+                }
+                let placeholder = Arc::new(CodebaseIndex::empty());
+                match shared.write() {
+                    Ok(mut g) => *g = placeholder,
+                    Err(p) => *p.into_inner() = placeholder,
+                }
+
                 match build_index(&path) {
                     Ok(fresh) => {
                         let new_arc = Arc::new(fresh);
@@ -2666,7 +2690,16 @@ pub fn spawn_mcp_watcher(
                         }
                     }
                     Err(e) => {
+                        // Must not leave the cell in `Building` — that state means
+                        // "retry shortly", and nothing is coming. Publish `Failed`
+                        // so tool calls get the reason instead of hanging on a
+                        // rebuild that already gave up.
                         eprintln!("cxpak: full rebuild after MCP overflow failed: {e}");
+                        let msg = format!("index rebuild after watcher overflow failed: {e}");
+                        match readiness.write() {
+                            Ok(mut g) => *g = IndexReadiness::Failed(msg),
+                            Err(p) => *p.into_inner() = IndexReadiness::Failed(msg),
+                        }
                     }
                 }
                 continue;
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -1414,11 +1414,37 @@ pub fn run(
             }
         };
 
+        // Bounded + coalesced + windowed debounce: collect_debounced blocks up
+        // to 1 s for the first event, then drains in 50 ms slices until the
+        // queue has been quiet for 400 ms (capped at 2 s total), and returns a
+        // de-duplicated batch.  If the bounded channel was overwhelmed and
+        // non-noise events were dropped, take_overflow() returns true and we
+        // rebuild the index from scratch (build_index) rather than trusting the
+        // lossy delta, swapping the fresh index into the shared handle.
         loop {
-            if let Some(first) = watcher.recv_timeout(Duration::from_secs(1)) {
-                let mut changes = vec![first];
-                std::thread::sleep(Duration::from_millis(50));
-                changes.extend(watcher.drain());
+            let changes = watcher.collect_debounced(
+                Duration::from_secs(1),
+                Duration::from_millis(400),
+            );
+            if watcher.take_overflow() {
+                eprintln!(
+                    "cxpak: watcher event buffer overflowed — \
+                     delta may be stale; triggering full rebuild"
+                );
+                // Re-index from scratch by swapping in a freshly built index.
+                match build_index(&watcher_path) {
+                    Ok(fresh) => {
+                        let new_arc = Arc::new(fresh);
+                        match watcher_index.write() {
+                            Ok(mut g) => *g = Arc::clone(&new_arc),
+                            Err(p) => *p.into_inner() = Arc::clone(&new_arc),
+                        }
+                    }
+                    Err(e) => {
+                        eprintln!("cxpak: full rebuild after overflow failed: {e}");
+                    }
+                }
+            } else if !changes.is_empty() {
                 process_watcher_changes(&changes, &watcher_path, &watcher_index);
             }
         }
@@ -2614,13 +2640,46 @@ pub fn spawn_mcp_watcher(
             }
         };
 
+        // Bounded + coalesced + windowed debounce: collect_debounced blocks up
+        // to 1 s for the first event (keeping shutdown polling responsive at
+        // ~1 Hz), then drains in 50 ms slices until the queue has been quiet
+        // for 400 ms (capped at 2 s total), and returns a de-duplicated batch.
+        // If the bounded channel overflowed, take_overflow() returns true and
+        // we fall back to a full rescan rather than trusting a lossy delta.
         while !shutdown.load(Ordering::Relaxed) {
-            let Some(first) = watcher.recv_timeout(Duration::from_secs(1)) else {
+            let changes = watcher.collect_debounced(
+                Duration::from_secs(1),
+                Duration::from_millis(400),
+            );
+
+            if watcher.take_overflow() {
+                eprintln!(
+                    "cxpak: MCP watcher event buffer overflowed — \
+                     delta may be stale; triggering full rebuild"
+                );
+                match build_index(&path) {
+                    Ok(fresh) => {
+                        let new_arc = Arc::new(fresh);
+                        match shared.write() {
+                            Ok(mut g) => *g = Arc::clone(&new_arc),
+                            Err(p) => *p.into_inner() = Arc::clone(&new_arc),
+                        }
+                        match readiness.write() {
+                            Ok(mut g) => *g = IndexReadiness::Ready(new_arc),
+                            Err(p) => *p.into_inner() = IndexReadiness::Ready(new_arc),
+                        }
+                    }
+                    Err(e) => {
+                        eprintln!("cxpak: full rebuild after MCP overflow failed: {e}");
+                    }
+                }
                 continue;
-            };
-            let mut changes = vec![first];
-            std::thread::sleep(Duration::from_millis(50));
-            changes.extend(watcher.drain());
+            }
+
+            if changes.is_empty() {
+                continue;
+            }
+
             // Publish the exact `Arc` the rebuild swapped into `shared` (already
             // embeddings-free per ADR-0200). `None` = nothing relevant changed,
             // so the prior `readiness` cell stands (issue #47 / ADR-0205).
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -1422,10 +1422,8 @@ pub fn run(
         // rebuild the index from scratch (build_index) rather than trusting the
         // lossy delta, swapping the fresh index into the shared handle.
         loop {
-            let changes = watcher.collect_debounced(
-                Duration::from_secs(1),
-                Duration::from_millis(400),
-            );
+            let changes =
+                watcher.collect_debounced(Duration::from_secs(1), Duration::from_millis(400));
             if watcher.take_overflow() {
                 eprintln!(
                     "cxpak: watcher event buffer overflowed — \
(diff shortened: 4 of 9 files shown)


cefothe (commented):
## Review: APPROVE ✅ (self-authored PR — posted as comment; GitHub blocks self-approval)

Well-scoped, thoroughly documented perf PR. All 6 CI checks pass (fmt, clippy, msrv, test, bench, coverage). Composite quality: **9.7/10**.

### Phase 0 — Security: CLEAN
- No secrets. Only URL is the standard crates.io registry in `Cargo.lock`.
- `tikv-jemallocator 0.6` — legitimate, checksum-pinned, `cfg`-gated off MSVC.
- `#[export_name = "malloc_conf"]` is correct jemalloc tuning, not a backdoor.

### Verified correctness
- **Noise/BUILTIN_IGNORES parity**: every `NOISE_COMPONENT` is present in `BUILTIN_IGNORES` (`.git` handled by `classify_changes`); pre-filter is a strict subset, enforced by `test_noise_components_are_rejected_by_classify_changes`.
- **Relativization guard** (`watcher.rs:325`): pre-filter strips the watch root before the noise check, so a noise-named ancestor (`/opt/build/repo`) can't blind the watcher. Covered by `test_noise_ancestor_does_not_blind_watcher`.
- **Overflow → full rebuild**: coalesced deltas correctly discarded in favor of `build_index`; poisoned-lock recovery present on all index swaps.

### Non-blocking findings
- **F1 (minor)**: the overflow→rebuild block is duplicated between `serve.rs:116-127` and `serve.rs:152-167`, differing only in the extra `readiness` write. Consider extracting a helper (CLAUDE.md: no code duplication).
- **F2 (minor)**: `test_coalescing_deduplicates_same_path_same_kind` asserts on a real FS write + 200ms sleep — mildly timing-dependent, could flake under CI load. The synthetic coalescing-map assertion below it is the deterministic part.
- **F3 (info)**: `muzzy_decay_ms:0` trades syscall/CPU churn for a tighter RSS ceiling — intentional and documented.

F1–F2 are optional polish; can land as follow-up or a quick amend. No security, correctness, or resource-safety blockers.

lyubomir-bozhinov (dismissed):
I ran the before/after measurement this PR ships `scripts/measure_rss.sh` for but does not report. Three paired experiments, both arms built `--release` from this branch and from `origin/main` (both report `cxpak 3.1.3`; only the PR binary contains `dirty_decay_ms`), same machine, same fixture, run back to back.

This matters more than usual here because **#48 shipped with a green memory-ceiling test and the bug came back anyway**. A third fix deserves a reproduction, not just a passing test.

**Short version:** the debounce/pre-filter half is a real and worthwhile improvement. The jemalloc half is not supported by what I can measure, and — the important part — **on the repo size where the 40 GB reports actually live, this PR does not help, and it adds a new multi-GB transient on exactly the wrong code path.**

---

## Experiment 1 — 302-file repo, 700 noise + 25 source writes per 1 s wave

RSS in MB, sampled 1 Hz, `SETTLE 60s / STORM 180s / QUIET 180s`.

| phase | main (3.1.3) | this PR |
|---|---|---|
| idle floor | 15 | 15 |
| storm median | **88** | **55** |
| storm max | 92 | 83 |
| quiet median | **15** | **15** |
| re-index batches | 179 | 168 |

## Experiment 2 — same repo, saturation (3 000 noise writes/wave, no inter-wave sleep)

| phase | main (3.1.3) | this PR |
|---|---|---|
| storm median | **91** | **62** |
| storm max | 99 | 84 |
| quiet median | **15** | **16** |
| **re-index batches** | **997** | **71** |

## Experiment 3 — 12 000-file repo, 25 source edits/s

| phase | main (3.1.3) | this PR |
|---|---|---|
| settle median | 3 003 | 4 517 |
| settle max | 3 901 | 6 098 |
| storm median | 2 706 | 2 970 |
| storm max | 4 266 | 5 029 |
| quiet median | 3 414 | 2 350 |
| quiet max | **5 232** | **3 544** |
| initial index completed? | **no, not in 6.5 min** | **no, not in 6.5 min** |

---

## 1. The stated premise for jemalloc does not reproduce

> the default system allocator … parks freed blocks in per-process free lists … so RSS ratchets to tens of GB and never returns

On macOS I cannot reproduce that. In experiments 1 and 2 the **unmodified 3.1.3 arm returns to its 15 MB idle floor** every time, at every event rate I could generate — including the saturation run where it did 997 full re-index batches in 150 s. Quiet-phase median is 15 MB on both arms; they are indistinguishable.

So the "never returns" behaviour isn't the system allocator on this platform, and Mechanism A is currently a fix for something I can't show is happening.

## 2. `background_thread:true` is rejected on macOS, loudly

Every launch of the PR binary prints to **stderr**, before anything else:

```
<jemalloc>: option background_thread currently supports pthread only
```

The code comment says it is "silently ignored on macOS" — it isn't silent. This is an MCP server on stdio; that line lands in Claude Code's logs on every start. Either `cfg`-gate the option to Linux or drop it, since macOS is the supported platform today.

## 3. The RSS win is the debounce doing 14× less work — which is good, but it's a different claim

Experiment 2 is the clean read: under **identical** churn, main did **997** re-index batches and this PR did **71**. That's 71 batches over 150 s — one per 2.1 s, i.e. pinned exactly at `DEBOUNCE_MAX_MS`.

So the lower RSS is mostly "rebuilt the index 14× less often", not "the allocator returned pages". That is a genuine improvement and I'd take it. Two asks:

- **Describe it as the debounce win**, so the next person doesn't inherit "jemalloc fixed it".
- **State the freshness trade.** Under sustained churn the index is now up to ~2 s + rebuild-time stale where it was ~50 ms. For an MCP index that's very likely the right call — but it's a user-visible behaviour change and it isn't in the PR description or an ADR.

## 4. At the size where the bug lives, this PR doesn't help

Experiment 3 is the one that matters, and it's the one that changed my mind. At 12 000 files **both arms sit in the 3–6 GB band and neither finishes its initial index in 6.5 minutes.**

Please read the base-vs-PR difference there as *noise*: one run per arm on a 24 GB machine that was already paging (`Pageouts: 12 544 657` ≈ 48 GB lifetime). I am **not** claiming this PR is worse. The robust claim is the one both columns agree on — **multi-GB, and unchanged by this PR.**

That's expected once you look at where the memory goes, which is not the watcher at all.

## 5. The actual mechanism looks like `build_index`, which is whole-repo-in-RAM

`src/commands/serve.rs:191`-ish holds, simultaneously:

- `content_map: HashMap<String, String>` — the **full source text of every file**
- `parse_results` — every extracted symbol for every file
- `files` — the scan result

…and only then hands all three to `CodebaseIndex::build_with_content`. Peak is the whole repo plus its whole parse, live at once. That's ~300 KB/file on my fixture, which is exactly the 3–5 GB observed at 12 000 files.

This is consistent with **#47**'s diagnosis ("two full CodebaseIndex deep clones per watcher batch"): clone cost tracks **index** size, and index size tracks **repo** size. It also explains why the reports are 40+ GB on real repos while a 302-file fixture caps at 99 MB no matter how hard I hammer it — and why the growth users see is **spiky rather than monotonic** (I sampled a long-lived server at 1156 → 20 → 16 → 461 → 16 MB earlier).

Nothing in this PR touches that path.

## 6. Blocking concern: the new overflow handler re-opens #47's shape

```rust
if watcher.take_overflow() {
    match build_index(&path) {
        Ok(fresh) => {
            let new_arc = Arc::new(fresh);
            match shared.write() { Ok(mut g) => *g = Arc::clone(&new_arc), ... }
```

`build_index` runs to completion — allocating a **complete second index** — while the old one is still held live by `shared` and `readiness`. On experiment 3's repo that's a transient ~2× of a structure that is already multi-GB, on the watcher thread, fired **precisely when the process is under an event storm**, i.e. when memory pressure is already worst.

Both call sites do this (`run` at ~1414, `spawn_mcp_watcher` at ~2638). Overflow is rare by design, so this won't show up in normal use — it will show up in exactly the pathological case the PR exists to fix. Given #47 was *"two full index clones per batch"*, putting a full index build back on the watcher thread deserves a deliberate decision rather than falling out of the error path.

Options: drop the old `Arc` before building; bound the rebuild (skip if one is in flight); or reuse the incremental path with a dirty-everything marker.

## 7. Three of the new tests can't fail

`test_bounded_channel_overflow_sets_flag` never constructs a `FileWatcher`. It builds its own `sync_channel`, its own `overflow`/`drop_count` atomics, fills it, asserts `TrySendError::Full`, sets its own flag, then asserts its own flag. It would still pass if `take_overflow()` were `fn take_overflow(&self) -> bool { false }`.

Same shape in the second half of `test_coalescing_deduplicates_same_path_same_kind` and in `test_coalescing_keeps_distinct_kinds`: both build a local `HashMap` and assert on it.

Net effect: **`collect_debounced` and `take_overflow` — the two functions this PR adds and the serve loops depend on — have no direct coverage.** A real test would drive a `FileWatcher` and assert on what `collect_debounced` returns.

I'd single this out because it is the same failure mode as #48: a green suite next to a live bug. `test_noise_ancestor_does_not_blind_watcher` and `test_noise_components_are_rejected_by_classify_changes`, by contrast, are exactly right — they exercise the real thing and encode a real invariant.

---

## What I'd suggest

Land the debounce/pre-filter half — it's a solid 14×-fewer-rebuilds win and the noise pre-filter with its `classify_changes` parity test is well built.

Before it goes in: fix the overflow double-allocation (**6**), give `collect_debounced`/`take_overflow` tests that can fail (**7**), and either drop `background_thread` on macOS or accept the stderr line knowingly (**2**).

And I'd split the jemalloc change out. It isn't supported by anything I can measure, and shipping it alongside the debounce is what would let "jemalloc capped our RSS" become received wisdom when the data says the debounce did it.

The 40 GB issue itself I think needs a separate ticket against `build_index`'s whole-repo-in-RAM shape — I've cross-linked the #47 → #48 → 3.1.3 → #51 chain on #51. Happy to file that one and hand over the fixture; the harness is three scripts and reproduces the multi-GB band in about six minutes.

Marking request-changes for **6** and **7** specifically — the rest is discussion, and the core of this PR is worth landing.

lyubomir-bozhinov:
Follow-up to my review: I filed the root-cause ticket I offered there — **#53**, `build_index` whole-repo-in-RAM (5.2 GB peak, never completes, on a 12k-file repo; same band on this branch). The fixture generator and harness are in it.

Also posted live evidence on **#51**: a 1 Hz sampler caught a long-lived server going **18 MB -> 1190 MB in 3 seconds**, decaying over ~2 minutes, then re-spiking — with four concurrent servers peaking at ~3.0 GB combined. That decay profile is a large short-lived structure being built and released, which is why I do not think the allocator is the lever.

None of that diminishes the debounce work here — 997 -> 71 rebuilds under identical churn is a real win and I would like to see it land. The two blocking items in my review stand: the overflow handler building a full second index while the old one is live (**6**), and `collect_debounced` / `take_overflow` having no test that can fail (**7**).

nikimilenkov:
## Issues found — reading this PR against the #47 → #48 → #51 → #53 chain

Independent pass over the full diff, focused on surfacing problems (not prescribing changes). These extend @lyubomir-bozhinov's review with what the cross-linked tickets add.

**I1 — The new tests for this PR's own functions cannot fail.** `test_bounded_channel_overflow_sets_flag` builds its own `sync_channel` + atomics and never constructs a `FileWatcher`; both `test_coalescing_*` assert on a locally-built `HashMap`, not on `collect_debounced`. So `collect_debounced` and `take_overflow` — the two functions this PR adds, which both serve loops depend on — have no coverage that can go red. This is the same failure mode that already bit this bug: #47 → #48 landed a fix **with a passing P1 memory-ceiling test**, and #51 reports the bug **back in 3.1.2 and 3.1.3 anyway**. A green suite has demonstrably not been evidence here.

**I2 — The stated jemalloc premise is contradicted by production data.** Mechanism A's rationale is "freed pages are never returned to the OS, so RSS ratchets up and stays." #51's own 1 Hz samples show RSS **collapsing back to a ~16–20 MB baseline repeatedly** (`1156 → 20 → 16 → 461 → 16 MB`). Pages never returned cannot fall back to baseline, so the field signature is the opposite of the premise. Experiment 2 (identical churn: main 997 rebuilds vs this PR 71) indicates the measured RSS drop is the **debounce doing 14× less work**, not jemalloc. The jemalloc change is bundled with the debounce under a claim the data doesn't support.

**I3 — `background_thread:true` prints to stderr on macOS every launch.** `<jemalloc>: option background_thread currently supports pthread only` lands on stdout/stderr of an MCP server on stdio, on the supported dev platform, on every start. The code comment calls this "silently ignored" — it isn't silent.

**I4 — The "40 GB" is plausibly a fleet aggregate, unaddressed here.** #51 caught **five `cxpak serve --mcp` alive at once, two >2 days old.** Orphaned long-lived MCP servers each spiking multi-GB sum to the reported figure — a process-lifecycle problem that a per-process allocator/watcher change does not touch.

**I5 — The overflow fallback calls a measured-to-hang operation on the watcher thread.** #53 measures `build_index` at **5.2 GB peak, not completing within 6.5 min on a 12k-file repo.** This PR's overflow path invokes exactly that `build_index` **on the watcher thread under an event storm** (`serve.rs:1433` in `run`, `serve.rs:2656` in `spawn_mcp_watcher`), while the old index is still live in `shared` (and `readiness`). It can stall the freshness loop and double peak memory precisely in the pathological case this PR targets — the shape #47 was closed to remove.

**I6 — "Fixes the 40 GB" is contradicted by this repo's own open tickets.** The measured ceiling is #53 (`build_index` whole-repo-in-RAM), which this PR does not touch; #51 remains open across two releases.

**I7 (minor) — Doc/behavior inaccuracies.** `DEBOUNCE_SLICE_MS`'s comment at `watcher.rs:18` calls it the "inter-batch quiet period" — that's the `window` param (400 ms); the constant is the 50 ms poll slice. Separately, `collect_debounced` now imposes a ~400 ms floor before returning even for a single isolated edit (was a fixed 50 ms) — an 8× steady-state freshness change not noted in the description or an ADR.

For balance: the debounce/pre-filter mechanics are sound and well-tested at the invariant level — `test_noise_components_are_rejected_by_classify_changes` and `test_noise_ancestor_does_not_blind_watcher` exercise the real thing. The issues above concern the jemalloc half, the untestable overflow/debounce tests, and the "fixes the 40 GB" framing.

lyubomir-bozhinov:
I've pushed four commits on top of your work rather than handing the review back — you're busy, and the investigation turned up something big enough that describing it would have been slower than fixing it. **Additive only: no force-push, your commits are untouched.** I also merged `main` in to resolve the Cargo.lock overlap with #52.

**I have to correct my own review first.** I told you the jemalloc half wasn't supported by measurement and the debounce half was doing the real work. The debounce part stands. But I had the *root cause* wrong, and so did #47 — and that is why this bug keeps coming back.

## What it actually is

`intelligence::co_change::build_co_changes` expands each commit into every pair of its changed files:

```rust
for i in 0..files.len() {
    for j in (i + 1)..files.len() {
        let a = files[i].clone();   // two heap Strings
        let b = files[j].clone();   // per pair
        pair_data.entry(key)...     // HashMap<(String, String), _>
```

`F*(F-1)/2` pairs for a commit touching `F` files, allocated in full **before** any threshold pruning. One mass-touch commit is enough to eat the machine — initial import, formatter sweep, generated-code or vendored-dependency refresh, squash-merge of a long branch. A 12 000-file commit expands to **71 994 000 pairs**.

That single fact explains everything that was confusing about #51:

- **Why the growth is spiky, not monotonic** — it's a large transient that then frees. It matches the live trace I posted (18 MB → 1190 MB in 3 s, decay over ~2 min, re-spike).
- **Why it tracks repo size, not event rate** — I could not reproduce it on a 302-file repo at *any* event rate, including 997 full re-index batches in 150 s.
- **Why watcher-level fixes kept not holding** — #48 removed a multiplier (two deep clones per batch), but not the base cost of the thing being multiplied. 3.1.3 genuinely contains #48's fix, and the bug came back anyway.

## How I found it

Phase-resolved RSS probes through the whole index build, then a stack sample taken at 3 290 MB:

```
run_mcp → build_index_with_workspace → build_convention_profile
        → extract_git_health → build_co_changes      ← 6/6 samples
```

Every other phase of the build stayed under 500 MB and scaled **linearly** (128 MB @2k files, 202 MB @4k, 340 MB @8k, 413 MB @12k). The blowup is entirely in co-change mining.

Worth saying plainly: my first guess was `build_with_content`'s content maps, and the probes showed that was wrong — it's ~90 MB of the total. Measuring instead of reasoning is the only reason this landed on the right function.

## The fix

Two bounds plus interning (`127ae1a`):

- **`MAX_FILES_PER_COMMIT` (200)** — skip mass-touch commits. This *improves* the signal as well as the cost: a commit touching thousands of files links every file to every other file, which is not evidence that any particular pair is coupled. It's noise that also happens to dominate memory.
- **`MAX_TRACKED_PAIRS` (2 000 000)** — ceiling on distinct pairs so a long history can't rebuild the blowup one at-cap commit at a time. Existing pairs keep accumulating; only new ones are dropped.
- **Paths interned to `u32`** — the map holds two integers per entry instead of two heap `String`s. Ordering is still decided by comparing the *paths*, so emitted `(file_a, file_b)` ordering is byte-identical to before.

Neither bound truncates silently — both counts go to stderr. A silent cap reads as "mined everything" when it didn't, which is the trap #48's green ceiling test fell into.

## Measured

`cxpak serve --mcp` on a 12 000-file repo, release builds, same fixture, back to back:

| | peak RSS | index ready |
|---|---|---|
| `main` 3.1.3 | 5 036 MB | **never** (>242 s) |
| this PR before my commits | 7 235 MB | **never** (>242 s) |
| this PR after | **506 MB** | **yes — 16 s** |

10× lower peak, and "never finishes" becomes "ready in 16 seconds", with all 12 001 files indexed. On a normal repo (cxpak itself, 520 files) neither bound fires and behaviour is unchanged.

Note the middle row: this branch as it stood measured *worse* than main. I saw the same in an earlier run. I'd treat the exact numbers as noisy — single runs on a machine that was already paging — but the direction was consistent across both, so it's worth knowing.

## The three review findings

- **Finding 6** (`2dedc1b`) — the overflow handler built a full second index while the old one was still held by `shared` and `readiness`. Both references are now released first, publishing `Building` for the duration. The trade is availability for memory and it's the right way round: overflow means events were dropped, so the index has just been declared lossy — "indexing in progress, retry" is an honest answer to a question we can no longer answer correctly, and clients already retry on it (#19/ADR-0185). The error path publishes `Failed` rather than leaving the cell stuck on `Building`. MCP path only; the HTTP watcher has no readiness cell so it keeps its zero-downtime 2× swap by design.
- **Finding 7** (`323532c`) — replaced the three self-asserting tests with ones that drive a real `FileWatcher`. `with_bound` is the seam that makes overflow reachable without queueing 65 536 events.
- **Finding 2** (`c1973f5`) — `background_thread` scoped to Linux, so macOS stops printing the jemalloc warning to an MCP server's stderr on every launch.

**One of those new tests found a real latent bug.** `noise_paths_do_not_overflow_the_channel` failed: the pre-filter only runs when `strip_prefix(root)` succeeds, and event paths arrive fully resolved, so a relative root (`.`) or a symlinked one (macOS `/var` → `/private/var`) makes every strip miss and silently degrades the filter to a no-op — passing exactly the churn it exists to drop. That's issue #18's failure mode. **Not currently live** — both production callers canonicalize already, and `spawn_mcp_watcher` even has the comment explaining why — but the guarantee belonged inside `FileWatcher` rather than depending on caller discipline. It's there now.

## Gate

`cargo test`: **3 147 passed, 0 failed, 3 ignored** across 81 suites. `clippy --all-targets -D warnings` clean. `fmt --check` clean. Watcher module re-run 5× for flakiness — stable.

## Where that leaves the PR

I'd now say **land the whole thing**. Your debounce is a genuine 997 → 71 rebuild reduction, and with the co-change fix underneath it the numbers finally move in the right direction rather than sideways.

I'd still split the jemalloc change out, for the same reason as before but more so: the data now points somewhere specific, and shipping the allocator alongside it is how "jemalloc capped our RSS" becomes received wisdom when it was a quadratic in co-change mining. Your call — it's your PR and I've said my piece.

I've authored commits here, so **I can't approve this**. It needs your review.

Refs #53, #51, #47

PR #56, opened 2026-08-05
docs(#42): SECURITY.md advertised a sandbox for a loader that cannot run plugins

### LINKED ISSUE #42: WASM plugin subsystem is a non-functional stub, yet SECURITY.md advertises its sandbox
**Severity: Medium** (shipped stub + over-claimed security doc)

## Problem
The WASM plugin subsystem is non-functional: `plugin/loader.rs:122` — `load()` always returns `Err("WASM plugin loaded (N bytes) but guest function binding not yet implemented")`, and a test (`load_valid_wasm_..._returns_guest_function_binding_err`) asserts exactly that. No guest function can ever run. Yet `SECURITY.md:50-52` advertises "plugins run in a wasmtime sandbox with … memory and CPU (epoch-interruption) caps" as a security control, and the resource caps that do exist are porous (`table_growing` returns `Ok(true)` unconditionally; fuel disabled; the 1MB return cap is enforced after `serde_json` deserialization).

## Why it matters
A published v3.1.2 crate ships a placeholder implementation ("not yet implemented") — violating the project's own no-placeholders / implementation-depth standard — and SECURITY.md describes a sandbox for a feature that cannot execute. It is off-by-default (limited blast radius), but the doc over-claims a control that never runs, and the epoch-thread/capability wiring is dead complexity.

## Proposed fix
Either finish the WIT guest bridge or remove `plugin/loader.rs` + the SDK story until it's real; strike/qualify the SECURITY.md sandbox claim; before shipping `plugins`, bound tables, enable fuel, and check byte-length pre-deserialize.

## Acceptance criteria
- No shipped "not yet implemented" plugin path, or `plugins` is clearly documented as non-functional.
- SECURITY.md describes only controls that execute.

lyubomir-bozhinov:
Closes #42. Also carries A2 of Barnett-Studios/dotclaude#124 (plane + status line).

## The finding, re-verified on `main` at 3.1.4

`src/plugin/loader.rs:85-127` — `load()` ends unconditionally in:

```
Err("WASM plugin loaded ({} bytes) but guest function binding not yet implemented")
```

There is no WIT bridge. No guest function has ever been callable, and a test pins that error.

`SECURITY.md:50-53` nevertheless listed the wasmtime sandbox under **security-relevant surfaces**. That is the worst shape a security document can take — not an overstatement of a weak control, but a description of a control with nothing running behind it. A researcher reads it and calibrates against a boundary that has never been exercised.

## What changed

- **`SECURITY.md`** — the plugin bullet now says the loader is non-functional, with a dedicated section explaining why the wasmtime configuration should not be read as a control. It names the three gaps verified in the source: `table_growing` returns `Ok(true)` unconditionally (`loader.rs:33-39`), `consume_fuel(false)` (`:53`), and the 1 MiB return cap applied post-deserialization.
- **`README.md`** — "WASM Plugin SDK / Extend cxpak with custom analyzers and detectors" was the same claim aimed at users. Rewritten to state plainly that you cannot extend cxpak with a plugin today.
- The **"not a vulnerability"** bullet about an untrusted plugin accessing repo content is removed. It presumed execution. In its place: findings against the *loading* path — size cap, checksum comparison, compilation, instantiation — are explicitly in scope, because those steps do run.

## What this does not do

It does not build the sandbox, per the brief. Acceptance criterion 2 of #42 ("SECURITY.md describes only controls that execute") is met by correcting the doc; criterion 1 was already met by `Cargo.toml`, which excludes `plugins` from `default` with a comment saying why.

No behaviour change, no code touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@
 ![Downloads](https://img.shields.io/crates/d/cxpak)
 ![License](https://img.shields.io/badge/License-MIT-green.svg)
 
+**Context plane · Active** — under development; the surface still moves.
+See the [component map](https://github.com/Barnett-Studios) for how this fits the rest.
+
 **Spends CPU cycles so you don't spend tokens.**
 
 cxpak indexes your codebase using tree-sitter across 43 languages, builds a typed dependency graph, and produces token-budgeted context bundles that give LLMs a briefing packet instead of a flashlight in a dark room. It understands your code's architecture, conventions, risk profile, and data layer -- then packs exactly what the LLM needs, nothing more.
@@ -315,13 +318,19 @@ Minimal local config:
 { "embeddings": { "provider": "local" } }
 ```
 
-## WASM Plugin SDK
+## WASM plugin loader — a skeleton, not a working SDK
 
-Extend cxpak with custom analyzers and detectors:
+**You cannot extend cxpak with a WASM plugin today.** `PluginLoader::load()` verifies the module's
+checksum, compiles it, instantiates it, and then returns `guest function binding not yet
+implemented`. The WIT bridge that would make a guest function callable does not exist, so no plugin
+code has ever run.
 
-- Plugin manifest with SHA-256 checksum verification before WASM compilation
-- wasmtime sandbox: epoch interruption (CPU), 64 MB memory cap, capability enforcement
-- File pattern scoping and content access control
+The `plugins` feature is therefore excluded from `default` — a stock `cargo install cxpak` does not
+build it, and the plugin management commands are behind the same gate. What is built out is the
+loading path: a manifest with SHA-256 verification before compilation, a 10 MiB module cap, a 64 MiB
+memory limiter, and a 10 s epoch deadline. Treat none of it as a security boundary until the bridge
+lands and the gaps in [`SECURITY.md`](SECURITY.md#the-wasm-plugin-loader-does-not-run-plugins) are
+closed — [#42](https://github.com/Barnett-Studios/cxpak/issues/42).
 
 ## Workspace support
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -47,17 +47,46 @@ surfaces worth a security researcher's attention:
   to a running Postgres/MySQL over rustls (no OpenSSL), issues a **read-only**
   session, and **never logs or persists the DSN**. Off unless explicitly built
   with the feature.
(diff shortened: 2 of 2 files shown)


lyubomir-bozhinov:
**Approved.** Every specific in this diff verifies against `src/plugin/`, which is the standard a security-claim *retraction* has to meet — the previous text failed it, so I checked all nine.

| claim | source |
|---|---|
| `load()` returns `guest function binding not yet implemented` | `loader.rs:123` |
| checksum verified **before** compilation | `verify_checksum_bytes` at `:104`, `Module::new` at `:106` |
| …in constant time | `subtle::ConstantTimeEq` + a length pre-check, `:74-78` |
| 10 MiB module cap | `MAX_PLUGIN_BYTES = 10 * 1024 * 1024` `:7` |
| 64 MiB memory limiter | `MAX_PLUGIN_MEMORY = 64 * 1024 * 1024` `:11` |
| 10 s epoch deadline | `EPOCH_DEADLINE = 100` ticks × 100 ms `:14` |
| `table_growing` unbounded | `Ok(true)` unconditionally `:33-39` |
| fuel metering disabled | `config.consume_fuel(false)` `:53` |
| 1 MiB cap applied after deserialization | `security.rs:13` — and the function's own doc comment says so |
| a test asserts the error | `loader.rs:237-261` |
| `plugins` excluded from `default` | `Cargo.toml [features] default` — 43 `lang-*` + daemon/embeddings/lsp/visual, no `plugins` |

The 1 MiB point is the one I expected to have to argue with, and the source is ahead of you: `enforce_return_limit`'s doc comment already states it "runs on the host AFTER the `Vec<Finding>` has been deserialized". Your phrasing is sharper — *the allocation it is meant to bound has already happened* — and that is the reason it matters.

**"43 languages" checks out too**, which I mention because it is the kind of number that rots quietly: I counted the `lang-*` entries in `default` and got exactly 43.

## Why this one is worth the care

The old text listed the wasmtime config as an active control and put "running an untrusted plugin and having it access repository content" under *not a vulnerability* — an exclusion for a threat model that cannot exist, since no guest code can run. Replacing that with a scoped invitation (the loading path *is* in scope; guest-execution findings are not) is strictly better than deleting it, because it tells a researcher where to look rather than only where not to.

The retraction sentence — *"Earlier revisions of this document described the sandbox as an active control. That was wrong"* — is the right way to do this. A security doc that quietly stops claiming something leaves everyone who read the old version still believing it.

Issue #32, opened 2026-07-20
Default feature set is a kitchen sink (embeddings/lsp/visual compiled by cargo install)


lyubomir-bozhinov:
**Severity: High** (scope / DX / Unix philosophy)

## Problem
`Cargo.toml` `default` compiles 43 tree-sitter grammars **plus** `daemon`, `embeddings` (candle-core/nn/transformers — a full ML runtime, ~90MB weights), and `lsp` (and the visual dashboard pulls resvg). A plain `cargo install cxpak` therefore triggers multi-minute cold builds and hundreds of MB for a tool whose core value — index → graph → token-budgeted context — needs none of embeddings/lsp/visual.

## Why it matters
This is the first thing an OSS maintainer criticizes: a kitchen-sink binary (CLI + MCP + HTTP + LSP + WASM + daemon + D3 dashboard + ML embeddings). It undercuts the "spends CPU so you don't spend tokens" pitch with a heavyweight install and a huge dependency/attack surface, and slows adoption.

## Proposed fix
Ship a lean `default` (core + a sensible language subset). Make `embeddings`, `lsp`, `visual`, `daemon` opt-in features, or split `cxpak-visual`/`cxpak-lsp` into separate crates.

## Acceptance criteria
- `cargo install cxpak` with defaults does not compile candle/resvg/tower-lsp.
- Heavy capabilities are documented opt-in features.

---

## Update (2026-07-25): OpenKB / vectorless retrieval reinforces trimming embeddings
OpenKB (VectifyAI) does reasoning-based **vectorless** retrieval (PageIndex hierarchical tree; explicitly no embeddings) — and cxpak is already mostly vectorless (graph + PageRank + 7-signal lexical fusion). Per the audit the compiled-in embedding/candle path is **dead, unmeasured latency** (NF2). So the default-feature trim here should include **dropping the embedding default** unless it earns an evidence gate. PageIndex-style hierarchical indexing is worth evaluating for large files (also helps the O(N²) in #37). **Don't add a vector DB (LanceDB)** — the signal is we may need none.

PR #58, opened 2026-08-06
feat(#57): ship the official Rust client with the server, behind a non-default feature

### LINKED ISSUE #57: Ship an official Rust client with the server: absorb baseplate's rmcp client behind a non-default `client` feature
Step 1 of the `baseplate` fold ([baseplate#4](https://github.com/Barnett-Studios/baseplate/issues/4)). A client belongs with its server, and an official Rust client shipped *with* cxpak makes the component adoptable in a way a client living in an unrelated "substrate" crate never will.

## What moves

`baseplate/src/cxpak.rs` — 535 lines, the largest file in that crate:

- `trait CxpakClient` — the seam consumers depend on
- `RmcpCxpakClient` — the live child-process transport, with `SpawnBackoff` (3-strike) and the "indexing in progress" polling loop
- `RecordedCxpakClient` — the replay/fixture client, which is what every existing test actually uses
- the typed response structs: `Health`, `DeadCode`, `DeadSymbol`, `Predict`, `CallGraph`, `Verify`, `Violation`, `Architecture`, `ArchModule`, `SecuritySurface`

## Why this is not free, and what it should cost

**cxpak has zero rmcp references today** — `grep -rn rmcp src/ Cargo.toml` returns nothing. MCP is served by hand out of `src/commands/serve.rs`. So this move introduces **one** genuinely new dependency, `rmcp` (with `client` + `transport-child-process`), plus a wider tokio feature set than the current optional one.

> **Correction (#58 review).** This originally said *two* new dependencies, counting `async-trait`. It is not one: `async-trait v0.1.89` is already in `main`'s **default** tree transitively, at the identical version, so nothing extra is built for it — measured by diffing the two trees, not inferred. Same mistake as the `libc` one corrected earlier in this issue, and caught the same way: by checking each of the three separately instead of trusting the list I had written. Nothing about the gating or the acceptance criteria changes — AC1 is about `--no-default-features`, where async-trait is genuinely absent, and it still holds.

That lands directly on top of [#32](https://github.com/Barnett-Studios/cxpak/issues/32) — *"Default feature set is a kitchen sink (embeddings/lsp/visual compiled by `cargo install`)"*. Making the client a default feature would make that issue measurably worse for every user who wants an indexer and not an MCP client.

**So: a non-default `client` feature.** `cargo install cxpak` must not build rmcp. The acceptance criteria below pin that rather than leaving it to review.

## Acceptance

- [ ] `cargo tree -e normal --no-default-features` shows no `rmcp` and no `async-trait`
- [ ] `cargo build` (default features) does not compile the client module
- [ ] `cargo build --features client` does, and `RecordedCxpakClient` round-trips a recorded fixture
- [ ] a RED test committed **first** that fails against current `main` — the client type is simply absent there, so the test compiles only once the module exists
- [ ] the mutation is named in the commit body and confirmed to go red on revert
- [ ] ~~`CONTRACT.md` describes the new surface~~ **cxpak has no `CONTRACT.md`** — not on `main`, not on the branch. It is the one Rust component in the family without one, and this criterion was written as though it were not. Amended per the #58 review, which offered either option: **`README.md` describes the new surface**; the client is documented as a feature, not a default. Writing cxpak's first CONTRACT.md is a component-wide job and does not belong inside the PR that adds one module — filed separately.

## What this does not do

It does not remove anything from `baseplate`. That is step 2's problem, and `baseplate` keeps publishing until its consumers have moved — nothing is yanked, per baseplate#4.

## Licensing

Settled by the repo owner before any code moved: **cxpak becomes `MIT OR Apache-2.0`**, matching `attestr`, `commitward`, `abproof`, `cascadr` and `baseplate`. cxpak is currently the only Rust crate in the family that is MIT-only.

Dual-licensing is strictly additive — it removes no right an MIT recipient has today and adds an express patent grant for those who want one. It also means the moved code keeps the terms it was published under rather than being narrowed, so the move needs no relicensing of anyone's contribution.

Provenance was checked before the decision was taken: every one of the nine `baseplate/src/*.rs` files is single-author (three git identity strings, all the same person, traced back to the framework's first commit), with no fork PRs and no vendored third-party fragments. There is no third-party contributor whose permission would be required.

lyubomir-bozhinov:
Closes #57. Step 1 of the baseplate fold ([baseplate#4](https://github.com/Barnett-Studios/baseplate/issues/4)).

Moves `baseplate/src/cxpak.rs` (535 lines) to `src/client.rs`. A client belongs with its server, and an official Rust client shipped *with* cxpak makes the component adoptable in a way one living in an unrelated "substrate" crate never will.

## The feature boundary is the point, and it is verified not asserted

**cxpak had zero rmcp references before this** — MCP is served by hand out of `commands/serve.rs` — so this introduces `rmcp` and `async-trait` as genuinely new dependencies. That lands straight on [#32](https://github.com/Barnett-Studios/cxpak/issues/32), *"default feature set is a kitchen sink"*. So the client is **non-default**:

| | rmcp |
|---|---|
| `cargo tree -e normal --no-default-features` | absent |
| `cargo tree -e normal` (default) | 0 occurrences |
| `cargo tree -e normal --features client` | present |

`cargo install cxpak` stays an indexer.

## A correction to my own acceptance criteria

I wrote #57's acceptance list to require `libc` absent under `--no-default-features` alongside rmcp. **That was wrong, and I only caught it by checking all three separately instead of trusting the list I had written.** libc is already in the clean tree 8 times transitively (via `sha2`/`cpufeatures` and `fs2`) and is not a direct dependency without the feature. The `client` feature promotes it to direct for `client.rs:497`'s PID-based SIGKILL teardown; it adds nothing to the build graph. Issue updated.

## What the RED test pins, and a bug it prevented

Committed first, failing with `the package 'cxpak' does not contain this feature: client`. Mutation confirmed: reverting the module, its `lib.rs` declaration and the feature returns exactly that error.

**My first version of that test asserted an API I had not read.** I wrote it from how dotclaude-core's tests *construct* the client and inferred the rest, giving `call()` a `Result<String>` return over a `HashMap<(String,String),String>`. The real signature is `async fn call(&self, tool: &str, args: Value) -> Option<Value>` over `HashMap<String, Value>`.

That distinction is load-bearing: **`None` means the tool was unavailable, errored or degraded — never "the server looked and found nothing"**. Callers map `None` to a skipped observation rather than to a verdict. Had I implemented against my invented `Result` signature, the move would have erased the silent-false-negative guard the seam exists for. The corrected test pins it, plus `from_dir`'s missing-directory-is-an-error behaviour, which is the same distinction one layer up.

## Licence

**cxpak becomes `MIT OR Apache-2.0`** — the owner's decision, taken before any code moved. cxpak was the only Rust crate in the family that was MIT-only. Dual-licensing is strictly additive: it removes no right an MIT recipient has and adds an express patent grant.

Worth recording that the brief framing this work said *"baseplate is Apache-2.0, cxpak is MIT"* and posed the move as a relicensing problem. baseplate is **dual**, so a recipient could already take that file under MIT alone — there was never a legal blocker. Provenance was checked before asking anyway: all nine `baseplate/src/*.rs` files are single-author, traced to the framework's first commit, no fork PRs, no vendored fragments.

`LICENSE` → `LICENSE-MIT` via `git mv` (history preserved), `LICENSE-APACHE` added. Nothing yanked, no grant withdrawn.

## Not in scope

Nothing is removed from `baseplate` — that is step 2, and baseplate keeps publishing until its consumers have moved.

3/3 green under `--features client`; `cargo fmt --all -- --check` clean.


lyubomir-bozhinov:
## Blocking — the six green checks do not touch the 1,131 lines this PR ships

The feature gating is correct, the tests are good, and I reproduced every build-level criterion. The problem is that **nothing in CI ever compiles the client**, so the green rollup on this PR is evidence about the indexer, not about the thing being added.

Every job runs default features:

```
ci.yml:42  cargo clippy --all-targets -- -D warnings     # client not compiled → never linted
ci.yml:51  cargo test --verbose                          # client_feature_boundary → 0 tests
ci.yml:69  cargo test --features bench ...               # bench only
ci.yml:95  cargo build --verbose                         # default
```

Measured, not inferred:

```
$ cargo test --test client_feature_boundary        # default features, as CI runs it
running 0 tests
test result: ok. 0 passed; 0 failed; ...
```

`ok`, in green, having asserted nothing. With the feature it is a real suite:

```
$ cargo test --features client --test client_feature_boundary
test recorded_client_replays_a_committed_recording ... ok
test an_unrecorded_tool_is_a_miss_not_an_empty_success ... ok
test from_dir_rejects_a_missing_directory_rather_than_returning_an_empty_client ... ok
test result: ok. 3 passed; 0 failed
```

So the RED-first discipline was real and the tests are worth having — and after merge nothing will ever run them again. The next PR that breaks `RmcpCxpakClient` goes green.

This is the class corpus#28 is currently documenting one repo over: *a check that passes for the wrong reason, whose only signal is the shape of a green result.* Here it is arriving in the same week, in the same family, on a PR whose whole point is that the client is now a first-class shipped artifact.

**Fix:** add the feature to the jobs that should cover it — `cargo test --features client` and `cargo clippy --features client --all-targets`. I ran clippy under the feature myself and it is **clean, zero warnings**, so this costs a CI minute and no code changes. Same PR, please: a shipped surface that CI cannot see is not shipped, it is vendored.

## Blocking — acceptance criterion 6 is unmet, and the file it names does not exist

> - [ ] `CONTRACT.md` describes the new surface; the client is documented as a feature, not a default

**cxpak has no `CONTRACT.md`** — not on this branch, not on `main` (`git ls-tree origin/main` finds nothing). Every other Rust component in the family ships one; cxpak is the exception, and the criterion was written as though it were not.

The README is in the changed-file list, so I checked what it got. Four lines, all licensing:

```
-![License](https://img.shields.io/badge/License-MIT-green.svg)
+![License](https://img.shields.io/badge/license-MIT%20OR%20Apache--2.0-blue.svg)
-MIT
+Licensed under either of [MIT](LICENSE-MIT) or [Apache-2.0](LICENSE-APACHE) at your option. …
```

No mention of the `client` feature, no `CxpakClient`, no usage example, nothing about `features = ["client"]`. The only description of the new public API anywhere is a Cargo.toml comment and the rustdoc on `pub mod client`.

The issue's own framing is that shipping the client here *"makes the component adoptable in a way a client living in an unrelated substrate crate never will."* Adoptability is the thing documentation delivers. A user reading the README today cannot discover that cxpak has a Rust client at all.

Either create the `CONTRACT.md` the criterion assumes and describe the surface there, or amend the criterion and document it in the README — but one of the two, in this PR, since the criterion is checked here.

---

## Verified and correct

**AC1 — passes, measured both ways.**

```
cargo tree -e normal --no-default-features   → rmcp: 0, async-trait: 0
cargo tree -e normal                         → rmcp: 0
```

`cargo install cxpak` will not build rmcp. The stated motivation holds.

**AC2 — the gate is real.** `#[cfg(feature = "client")] pub mod client;` in `src/lib.rs`, and `client = ["dep:rmcp", "dep:async-trait", "dep:tokio", "dep:libc"]` is absent from `default`. Confirmed against the default tree above rather than by reading the manifest alone.

**AC3 — passes**, including `recorded_client_replays_a_committed_recording`, which is the fixture round-trip the criterion asks for.

**AC4 — RED first, genuinely.** `f74ebc5 test(#57): RED — cxpak ships no Rust client` precedes `0dfd10d feat(#57): ship the official Rust client…`. The test header states the RED state precisely: against `main`, `cargo test --features client` fails with *"the package 'cxpak' does not contain this feature: client"*.

**AC5 — the best-evidenced mutation note I have reviewed in this fleet.** The commit body names the revert, the resulting error, and that it matches the state the test was committed in. It also records the three tree checks separately rather than as one claim.

**Licensing is complete and consistent.** `license = "MIT OR Apache-2.0"` in Cargo.toml; `LICENSE-MIT` (1,074 B) and `LICENSE-APACHE` (10,835 B) are the standard texts; the old single `LICENSE` is gone; badge and README section both updated. This matches attestr, commitward, abproof, cascadr and baseplate, and it is strictly additive as the issue argues.

**Clippy is clean under the feature** — I ran `cargo clippy --features client --all-targets` and it exits 0 with no warnings. The code is fine; it is the coverage that is missing.

### One precision note, non-blocking

The commit body says the move *"introduces rmcp and async-trait as genuinely new dependencies."* True of **direct** dependencies. Not true of what gets built: `async-trait v0.1.89` is already in `main`'s default tree, transitively and at the identical version — I diffed the two trees specifically to check this.

Flagging it only because you caught precisely this distinction for `libc` in the same commit — *"it is NOT a new dependency: the clean tree already carries it 8 times transitively… which I only noticed by checking each of the three separately rather than trusting the list I had written"* — and then corrected the issue. `async-trait` earns the same correction, for the same reason. Nothing about the gating or AC1 changes; the behaviour is right either way.

I will re-review as soon as the head moves, and re-check only CI coverage and the docs criterion.

lyubomir-bozhinov:
## Both findings fixed. Head is `ad44672`

### CI now compiles the client — and the gate is checked in both directions

Reproduced yours first, verbatim: `cargo test --test client_feature_boundary` → **0 passed**,
green, having asserted nothing.

`--features client` on clippy and test. Features are additive, so both jobs still cover
everything the default set covered and additionally compile `src/client.rs`:

```
cargo clippy --all-targets --features client -- -D warnings   clean, 0 warnings
cargo test --features client                                  3154 passed, 3 ignored
cargo test --features client --test client_feature_boundary   3 passed  (was 0)
```

**Your finding names one direction of the gate; there are two, and the other was equally
unwatched.** Direction 1: the client must compile and run. Direction 2: `rmcp` must stay *out*
of the default build — the property that makes `cargo install cxpak` an indexer and not an MCP
client (#32). Nothing checked that either, and direction 1 on its own is satisfied by deleting
the gate entirely. So the new step asserts both, which also means it cannot pass because the grep
stopped matching:

```
rmcp absent from `cargo tree -e normal`                   → else fail: the gate leaks
rmcp present in `cargo tree -e normal --features client`  → else fail: the probe is broken
```

Mutation: add `"client"` to `Cargo.toml`'s `default` list → red on the first assertion. Verified,
not asserted.

**Your precision note is right and I re-measured it.** `async-trait v0.1.89` is in the default
tree transitively; only `rmcp` is genuinely new to what gets built. The gate deliberately does not
assert async-trait, and #57's body now carries the correction the same way its `libc` claim did.
AC1 is unaffected — it is scoped to `--no-default-features`, where async-trait is genuinely
absent.

### AC6 — amended to README, and the missing file filed

Took your second option. `README.md` gains a **Rust client (non-default `client` feature)**
section: the feature line, `CxpakClient` with the `None`-means-miss contract stated as the
contract rather than a footnote, `RmcpCxpakClient`'s lazy spawn / backoff / nulled child stderr,
and `RecordedCxpakClient` + `from_dir` for consumers' own tests including why a missing directory
is an error.

#57's criterion 6 is struck through and amended in place, with the reason, rather than quietly
re-scoped. Writing cxpak's first `CONTRACT.md` is a component-wide job — it is the only Rust
component in the family without one, and it has the largest surface (CLI, MCP, HTTP, LSP, now a
client) — so it is [#59](https://github.com/Barnett-Studios/cxpak/issues/59), not a section
bolted onto the PR that adds one module.

**One doc claim was wrong before it was written.** I had drafted the install line as
`default-features = false, features = ["client"]`. It does not build — 4 errors, `daemon` gates
`commands::serve`. What the README documents is what compiled: `features = ["client"]`, and
`["client", "daemon"]` for the lean path.

### Scope, stated rather than left to inference

Coverage still runs on its default-feature list, so `src/client.rs` is now **linted and tested
but not in the 90% tarpaulin denominator**. Adding it there is a different decision — 22 KB of
new code against a hard threshold — and I would rather you see it named than discover it.

That closes all five. corpus#28 (18 confirmed; 220 was wrong the other way — it is 225),
cordon#10 (claim narrowed, boundary pinned by a test, cordon#11 holds the widen-the-mount trade),
baseplate#11 (footer plus two louder copies above the fold), dotclaude#129, and this one. All
five heads have moved.

lyubomir-bozhinov:
## Re-checked `23b9d663...ad446723` — CI now compiles what it ships, merging

Two things were open: CI never built the client, and AC6.

**1. The feature is now genuinely covered.** I read the CI log for this head rather than the
rollup, because a green rollup was exactly the thing that proved nothing here:

- `client_feature_boundary` compiled with `--cfg 'feature="client"'`, `rmcp` linked in
- `running 3 tests` — not the `running 0 tests … ok` the default build produced
- clippy runs `--all-targets --features client`
- the gate step printed `default tree is rmcp-free; --features client pulls it in`

The both-directions design is the right call and the comment explains why: direction 1 alone is
satisfied by deleting the gate, and direction 2 alone is satisfied by the client never
compiling. Asserting the probe itself (`rmcp` absent under the feature ⇒ hard error, "the probe
is broken, not the gate") means this cannot pass by the grep quietly stopping to match. Leaving
`async-trait` out of the assertion is correct — v0.1.89 is already in the default tree
transitively, so asserting on it would assert a fact about the *baseline*, not about this gate.

**2. AC6.** The criterion was amended in #57 itself — README describes the surface, and cxpak's
first `CONTRACT.md` is a component-wide job filed as cxpak#59. That is the split I offered, the
README section landed, and the follow-up exists rather than being promised. Met.

The README section documents the thing most likely to be misused: `None` from `call()` is "the
tool was unavailable", not "the server found nothing", and `from_dir` on a missing directory is
an error rather than an empty client. A fixture set that silently reports a clean run over
nothing is the failure worth spending the paragraph on.

**Delay was mine** — my head-check reported UNCHANGED against an empty sha for ~20h.

Issue #64, opened 2026-08-14
The installed plugin's README advertises 13 MCP tools; the server advertises 5, and none of the 13 by those names


lyubomir-bozhinov:
Split out of #61 rather than folded into it: #61 is one runtime string naming an unadvertised
tool. This is the same defect one level up, in the artifact a user *installs*, and it is a
documentation rewrite rather than a two-line fix.

## Measured

`plugin/README.md:51` is headed **"MCP Tools (13)"** and lists the complete v2.x surface:

```
cxpak_auto_context · cxpak_overview · cxpak_trace · cxpak_stats · cxpak_diff
cxpak_context_for_task · cxpak_pack_context · cxpak_search · cxpak_blast_radius
cxpak_api_surface · cxpak_context_diff · cxpak_verify · cxpak_conventions
```

`tools/list` on `main` serves `capability::adapter::mcp_tool_schemas()`, which renders the intent
catalog — **five** tools: `cxpak_context`, `cxpak_graph`, `cxpak_data`, `cxpak_review`,
`cxpak_insight`, each selecting a capability via `op`.

**Not one of the 13 names appears in the advertised set.** Every row of that table names something
`tools/list` has not offered since the 3.0 consolidation (ADR-0182).

## Why it is worse here than in #61

`plugin/README.md` is the README of the **installed Claude Code plugin** (`plugin/.claude-plugin/`,
`plugin/.mcp.json`, `plugin/marketplace.json`). It is what a user reads to learn what they just
installed, and it describes a surface that no longer exists. The `focus` sentence above the table
("all tools support a `focus` path prefix parameter") is scoped to that same dead surface.

All 13 still *route* — `legacy_alias_to_op` maps 26 former names — so a permissive client that
copies a name out of the README succeeds, and nothing fails loudly. A client that builds its
callable set from `tools/list` gets none of them. That is the same asymmetry #61 turns on, and it
is why this survived: the doc is wrong in the one direction the alias hides.

## Also worth deciding, not just fixing

`README.md:187` says the v2.x names "remain callable as deprecated aliases **for one release**".
That horizon has passed — the aliases are still routing on the pinned 3.1.4, and a downstream
CLAUDE.md now records them as live-until-further-notice. Either the sentence is wrong or the
aliases are overdue; either way, a stated horizon nobody enforced is what let a 13-row table of
removed tools sit in the installed plugin.

That part is a deprecation decision and is the owner's, so it is named here rather than
implemented.

## Acceptance

[...]

Issue #45, opened 2026-07-20
MCP wire contract: #19 discriminator incomplete for real tool errors + no per-op schema/version


lyubomir-bozhinov:
**Severity: Medium** (MCP wire contract completeness)

## Problem
- The #19 retryable/terminal discriminator was added only to the **not-ready** path (`mcp_tool_status`, `serve.rs:4479`). Every genuine tool-execution failure — param validation, IO, git, render — routes through `mcp_tool_error` (`serve.rs:4452`) with `isError:true` and **no** `retryable` field. Clients still cannot distinguish a transient IO/git error (retryable) from a bad-param error (terminal) for real calls — the gap #19 set out to close, still open for actual tool errors.
- The MCP tool schema (`capability/adapter.rs:151`, `serve.rs:2777`) exposes only `op` with `additionalProperties:true`; per-op params are undocumented in the schema and **unvalidated** (a typo'd `tokens` silently defaults), and `serverInfo` carries only `CARGO_PKG_VERSION` — no contract/schema version a client can gate breaking op-param changes on.

## Why it matters
Consumers (dotclaude) can't reliably retry vs surface errors for real tool calls, and can't detect breaking wire-contract changes or catch mis-typed params — fragile integration.

## Proposed fix
Tag `mcp_tool_error` with a `retryable` classification for all failure classes; emit per-op `oneOf` input schemas and a contract-version field in `serverInfo`.

## Acceptance criteria
- Every MCP error carries a retryable/terminal classification.
- Clients can read a contract version and validate op params against a schema.

lyubomir-bozhinov:
**Premise correction (verified against source).** The "retryable-vs-terminal discriminator is still an open gap (#19)" framing is contradicted by **ADR-0201** (`docs/adrs/0201-mcp-tool-call-status-discriminator.md`, status **ACCEPTED**, 2026-07-16), which is the resolution of #19. It explicitly considered the fix this ticket proposes — **Option B**: tag every result/error including the shared `mcp_tool_error` helper — and **rejected** it: "tagging the shared helper labels IO/render/capability failures as `invalid_params` — a factual lie… it bakes a wrong contract into the exhaustive table test." Building/Failed already flow through `structuredContent{status,retryable}`; generic param/capability/IO errors deliberately do not.

[...]

PR #65, opened 2026-08-14
fix(#61): recommend a tool the server actually advertises

### LINKED ISSUE #61: cxpak_review op=review recommends calling cxpak_auto_context, which tools/list does not advertise
## Summary

`cxpak_review` with `op: "review"` tells the caller to invoke a tool the server does not advertise:

```json
{
  "modified_files": [], "new_files": [], "deleted_files": [],
  "recommendation": "No prior context snapshot. Call cxpak_auto_context first to establish a baseline."
}
```

`tools/list` on the same server (3.1.4) offers exactly five tools, and `cxpak_auto_context` is not
among them:

```
cxpak_context, cxpak_graph, cxpak_data, cxpak_review, cxpak_insight
```

## It works, until the client checks

`cxpak_auto_context` still routes — `serve.rs:2975` maps `"cxpak_auto_context" => "context"` — so a
permissive client that forwards any name succeeds. Verified over stdio MCP against the published
image: `cxpak_auto_context {task}` and `cxpak_context {op:"context", task}` return **byte-identical
5623-byte** responses.

The failure mode is a client that validates tool names against the advertised list before
dispatching — which is normal, and is what an agent harness does when it builds its callable set
from `tools/list`. For those, the recommendation is a dead end: the model is told to call a tool it
was never offered, and the only way to comply is to guess the modern name.

Still present on `main`: `src/auto_context/diff.rs:365`.

## Reproduction

```sh
mkdir r && cd r && git init -q . && echo 'x = 1' > a.py && git add -A && git commit -qm init
{ echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"qa","version":"1"}}}'
  echo '{"jsonrpc":"2.0","method":"notifications/initialized"}'; sleep 3
  echo '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"cxpak_review","arguments":{"op":"review"}}}'
  sleep 5
} | docker run --rm -i -v "$(pwd):/repo" ghcr.io/barnett-studios/cxpak:latest serve --mcp .
```

## Provenance

`Provenance: pre-existing — the string predates the 3.x rename; the tool was renamed to
`cxpak_context` and this guidance was not carried across. `serve.rs` retains the alias, so nothing
broke loudly enough to catch it.`

## Last known good

`never observed green` on 3.x — the advertised name and the recommended name have disagreed since
the rename.

## Suggested fix

One string: name the advertised tool and its op.

```
"No prior context snapshot. Call cxpak_context with op=\"context\" first to establish a baseline."
```

Worth a grep for the same drift elsewhere — any user-facing string naming a pre-3.x tool name has
the same shape, and the alias means none of them fail loudly.

## Not a defect (checked in the same pass, recorded so nobody re-files it)

The empty arrays above are **correct**. `op: "review"` compares against a prior *context snapshot*,
not against git, so with no baseline established every list is legitimately empty — and the
`recommendation` field says exactly that. I nearly filed the empty arrays as a silent-wrong-answer
defect after reading a truncated response that cut off the field explaining them. `op: "diff"` on
the same server, same repo, same moment correctly reports `changed_files: 3` with the real hunks
for a modify + add + delete change set.

Found by the QA loop, pass 8.

lyubomir-bozhinov:
Closes #61.

`cxpak_review op=review` recommended `cxpak_auto_context`. `tools/list` serves
`capability::adapter::mcp_tool_schemas()` — five intent-tools, and that is not one of them.

## Two strings, not one

The ticket names `diff.rs:365`. Grepping the module found a second with the same defect:

| site | before | after |
|---|---|---|
| `no_snapshot_recommendation` | `Call cxpak_auto_context first…` | `Call cxpak_context (op: "context") first…` |
| the delta text's closing line | `Re-run cxpak_auto_context…` | `Re-run cxpak_context (op: "context")…` |

Fixing only the one the ticket named would have left the other reachable on every non-empty diff —
the more common path of the two.

## Routing is untouched

The alias stays. `legacy_alias_to_op` still maps all 26 former names, and
`tests/mcp_stdio_framing.rs` / `tests/mcp_live_surface.rs` still exercise
`cxpak_auto_context` end to end. This PR changes what cxpak *tells a caller to do*, not what it
accepts — which is also why the bug survived: a permissive client that copied the name succeeded,
so nothing failed loudly.

## The assertions agreed with the defect

Three tests pinned it as `contains("cxpak_auto_context")` — `diff.rs` ×2 and `serve.rs` ×1. A
literal is exactly as stale as the string it guards, so the replacement derives its denominator
from `mcp_catalog_tools()`, the same source `tools/list` is built from:

```rust
fn assert_advertised_tools_only(recommendation: &str) {
    let advertised = mcp_catalog_tools().iter().map(|t| t.name.clone())…;
    assert!(!advertised.is_empty(), "…the catalog is not being read, so this proves nothing");
    // every cxpak_* token in the string must be in `advertised`
    // …and the string must name at least one, or it is guiding no one
}
```

**Three mutations, each red for a different reason:**

| mutation | failure |
|---|---|
| no-snapshot string back to the alias | `names "cxpak_auto_context", which tools/list does not advertise (advertised: ["cxpak_context", "cxpak_graph", "cxpak_data", "cxpak_review", "cxpak_insight"])` |
| re-run string back to the alias | same message, other call site |
| string names no cxpak tool at all | `names no cxpak tool at all, so it cannot be guiding anyone` |

The third is the one worth having: without it, deleting the tool name entirely satisfies "names
nothing unadvertised" while the recommendation guides no one. The empty-catalog guard is the same
idea one level down — a catalog that failed to load would otherwise make every assertion vacuous.

`serve.rs`'s MCP-level test now asserts both directions: the advertised name is present **and** the
deprecated alias is absent.

## Verification

- **2598 lib tests pass**, 0 failed (`cargo test --lib`).
- `cargo fmt` clean, `cargo clippy --all-targets -- -D warnings` clean.

## Filed, not fixed here

`plugin/README.md` is headed "MCP Tools (13)" and lists the entire pre-3.0 surface — **none** of
the 13 names is advertised by `tools/list`. Same defect class, but it is the README of the
*installed plugin* and the fix is a documentation rewrite plus a deprecation-horizon decision, so
it is **#64** with the measurement rather than scope creep here.

### DIFF
--- a/src/auto_context/diff.rs
+++ b/src/auto_context/diff.rs
@@ -361,9 +361,9 @@ pub fn no_snapshot_recommendation() -> ContextDelta {
         new_symbols: Vec::new(),
         removed_symbols: Vec::new(),
         graph_changes: Vec::new(),
-        recommendation:
-            "No prior context snapshot. Call cxpak_auto_context first to establish a baseline."
-                .to_string(),
+        recommendation: "No prior context snapshot. Call cxpak_context (op: \"context\") first to \
+             establish a baseline."
+            .to_string(),
     }
 }
 
@@ -441,7 +441,8 @@ fn build_recommendation(
     }
 
     parts.push(
-        "Re-run cxpak_auto_context to refresh the context with the latest changes.".to_string(),
+        "Re-run cxpak_context (op: \"context\") to refresh the context with the latest changes."
+            .to_string(),
     );
 
     parts.join(" ")
@@ -463,6 +464,47 @@ mod tests {
     // -----------------------------------------------------------------------
     // Helpers
     // -----------------------------------------------------------------------
+
+    /// Every `cxpak_*` name a recommendation puts in front of a caller must be one the MCP
+    /// server actually advertises.
+    ///
+    /// These strings told callers to `Call cxpak_auto_context` — a name `tools/list` has not
+    /// offered since the 3.0 intent-tool consolidation (ADR-0182). It still *routes*, via the
+    /// deprecated alias in `serve.rs`, so a permissive client never noticed; a client that
+    /// validates tool names against the advertised set before dispatching — which is what an
+    /// agent harness does — was told to call something it was never offered (cxpak#61).
+    ///
+    /// The denominator is `mcp_catalog_tools()`, the same source `tools/list` is built from,
+    /// so this cannot go stale against a future rename the way a literal assertion did.
+    fn assert_advertised_tools_only(recommendation: &str) {
+        let advertised: Vec<String> = crate::capability::adapter::mcp_catalog_tools()
+            .iter()
+            .map(|t| t.name.clone())
+            .collect();
+        assert!(
+            !advertised.is_empty(),
+            "no advertised tools found — the catalog is not being read, so this proves nothing"
+        );
+
+        let mut named = Vec::new();
+        for word in recommendation.split(|c: char| !(c.is_alphanumeric() || c == '_')) {
+            if word.starts_with("cxpak_") {
+                named.push(word.to_string());
+            }
+        }
+        assert!(
+            !named.is_empty(),
+            "recommendation names no cxpak tool at all, so it cannot be guiding anyone: \
+             {recommendation}"
+        );
+        for name in named {
+            assert!(
+                advertised.contains(&name),
+                "recommendation names {name:?}, which tools/list does not advertise \
+                 (advertised: {advertised:?}): {recommendation}"
+            );
+        }
+    }
 
     /// Build a minimal `CodebaseIndex` from a list of (relative_path, content)
     /// pairs without touching the filesystem.
@@ -742,11 +784,7 @@ mod tests {
             "Unexpected recommendation: {}",
             delta.recommendation
         );
-        assert!(
-            delta.recommendation.contains("cxpak_auto_context"),
-            "Recommendation should mention cxpak_auto_context: {}",
-            delta.recommendation
-        );
+        assert_advertised_tools_only(&delta.recommendation);
     }
 
     // -----------------------------------------------------------------------
@@ -807,10 +845,11 @@ mod tests {
             delta.recommendation
         );
         assert!(
-            delta.recommendation.contains("cxpak_auto_context"),
-            "Recommendation should suggest re-running cxpak_auto_context. Got: {}",
+            delta.recommendation.contains("Re-run"),
+            "Recommendation should suggest re-running the context tool. Got: {}",
             delta.recommendation
         );
+        assert_advertised_tools_only(&delta.recommendation);
     }
 
     // -----------------------------------------------------------------------
--- a/src/auto_context/diff.rs
+++ b/src/auto_context/diff.rs
@@ -465,47 +465,6 @@ mod tests {
     // Helpers
     // -----------------------------------------------------------------------
 
-    /// Every `cxpak_*` name a recommendation puts in front of a caller must be one the MCP
-    /// server actually advertises.
-    ///
-    /// These strings told callers to `Call cxpak_auto_context` — a name `tools/list` has not
-    /// offered since the 3.0 intent-tool consolidation (ADR-0182). It still *routes*, via the
-    /// deprecated alias in `serve.rs`, so a permissive client never noticed; a client that
-    /// validates tool names against the advertised set before dispatching — which is what an
-    /// agent harness does — was told to call something it was never offered (cxpak#61).
-    ///
-    /// The denominator is `mcp_catalog_tools()`, the same source `tools/list` is built from,
-    /// so this cannot go stale against a future rename the way a literal assertion did.
-    fn assert_advertised_tools_only(recommendation: &str) {
-        let advertised: Vec<String> = crate::capability::adapter::mcp_catalog_tools()
-            .iter()
-            .map(|t| t.name.clone())
-            .collect();
-        assert!(
-            !advertised.is_empty(),
-            "no advertised tools found — the catalog is not being read, so this proves nothing"
-        );
-
-        let mut named = Vec::new();
-        for word in recommendation.split(|c: char| !(c.is_alphanumeric() || c == '_')) {
-            if word.starts_with("cxpak_") {
-                named.push(word.to_string());
-            }
-        }
-        assert!(
-            !named.is_empty(),
-            "recommendation names no cxpak tool at all, so it cannot be guiding anyone: \
-             {recommendation}"
-        );
-        for name in named {
-            assert!(
-                advertised.contains(&name),
-                "recommendation names {name:?}, which tools/list does not advertise \
-                 (advertised: {advertised:?}): {recommendation}"
-            );
-        }
-    }
-
     /// Build a minimal `CodebaseIndex` from a list of (relative_path, content)
     /// pairs without touching the filesystem.
     fn make_index(files: &[(&str, &str)]) -> CodebaseIndex {
@@ -784,7 +743,19 @@ mod tests {
             "Unexpected recommendation: {}",
             delta.recommendation
         );
-        assert_advertised_tools_only(&delta.recommendation);
+        assert!(
+            delta
+                .recommendation
+                .contains("cxpak_context (op: \"context\")"),
+            "no-snapshot recommendation should name the advertised context tool and its op: {}",
+            delta.recommendation
+        );
+        assert!(
+            !delta.recommendation.contains("cxpak_auto_context"),
+            "recommendation still names the deprecated alias, which tools/list does not \
+             advertise: {}",
+            delta.recommendation
+        );
     }
 
     // -----------------------------------------------------------------------
@@ -849,7 +820,13 @@ mod tests {
             "Recommendation should suggest re-running the context tool. Got: {}",
             delta.recommendation
         );
-        assert_advertised_tools_only(&delta.recommendation);
+        assert!(
+            delta
+                .recommendation
+                .contains("cxpak_context (op: \"context\")"),
+            "re-run recommendation should name the advertised context tool and its op: {}",
+            delta.recommendation
+        );
     }
 
     // -----------------------------------------------------------------------
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -3508,7 +3508,9 @@ fn dispatch_capability_op(
                     "task": task,
                     "candidates": candidates,
                     "total_files_scored": all_scored.len(),
-                    "hint": "Review candidates and call cxpak_pack_context with selected paths, or use these as-is."
+                    // `cxpak_pack_context` is a 3.0 legacy alias, not a name `tools/list`
+                    // offers; the op lives under the `cxpak_context` intent-tool (cxpak#61).
+                    "hint": "Review candidates and call cxpak_context (op: \"pack_context\") with selected paths, or use these as-is."
                 }))
                 .unwrap_or_default(),
             )
@@ -4461,7 +4463,7 @@ fn dispatch_capability_op(
                 let _ = (visual_type, format, focus, symbol, files_arg);
                 mcp_tool_error(
                     id,
-                    "Error: cxpak_visual requires the 'visual' feature flag. Rebuild with: cargo build --features visual",
+                    "Error: cxpak_insight (op: \"visual\") requires the 'visual' feature flag. Rebuild with: cargo build --features visual",
                 )
             }
         }
@@ -4487,7 +4489,7 @@ fn dispatch_capability_op(
                 let _ = (focus, format);
                 mcp_tool_error(
                     id,
-                    "Error: cxpak_onboard requires the 'visual' feature flag. Rebuild with: cargo build --features visual",
+                    "Error: cxpak_insight (op: \"onboard\") requires the 'visual' feature flag. Rebuild with: cargo build --features visual",
                 )
             }
         }
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -9064,12 +9064,17 @@ mod tests {
         let response: Value = serde_json::from_slice(&output).unwrap();
         let content = response["result"]["content"][0]["text"].as_str().unwrap();
         let result: Value = serde_json::from_str(content).unwrap();
+        let rec = result["recommendation"].as_str().unwrap();
+        // The advertised name, not the deprecated alias: this assertion is what let the
+        // recommendation keep naming a tool `tools/list` stopped offering in 3.0 (cxpak#61).
         assert!(
-            result["recommendation"]
-                .as_str()
-                .unwrap()
-                .contains("cxpak_auto_context"),
-            "no-snapshot recommendation should mention cxpak_auto_context"
+            rec.contains("cxpak_context"),
+            "no-snapshot recommendation should name the advertised context tool: {rec}"
+        );
+        assert!(
+            !rec.contains("cxpak_auto_context"),
+            "recommendation still names the deprecated alias, which tools/list does not \
+             advertise: {rec}"
         );
     }
 
--- a/tests/advertised_tool_names.rs
+++ b/tests/advertised_tool_names.rs
@@ -1,0 +1,198 @@
+//! Every `cxpak_*` name cxpak puts in front of a caller must be one the MCP
+//! server actually advertises, and every `(op: "...")` beside it must be an op
+//! that tool hosts.
+//!
+//! cxpak#61: a `no prior snapshot` recommendation told callers to
+//! `Call cxpak_auto_context` — a name `tools/list` has not offered since the
+//! 3.0 intent-tool consolidation (ADR-0182). It still *routes*, via the
+//! deprecated alias table in `serve.rs`, so a permissive client never noticed;
+//! a client that validates tool names against the advertised set before
+//! dispatching — which is what an agent harness does — was told to call
+//! something it was never offered.
+//!
+//! The first fix corrected the two strings the ticket named. That is a
+//! hand-written denominator: a third site, the `context_for_task` hint in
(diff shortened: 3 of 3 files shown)


lyubomir-bozhinov:
## Review — verified against source at `51352c84`. All three mutations reproduce, the vacuity guard is real, and the enumeration of sites is one short. Blocking on the third one.

### The guard, measured

Positive control before any mutant: the three tests selected by name give `3 passed; 2598 filtered out`, so the filter is really running them.

| mutation | measured |
|---|---|
| HEAD, unmodified | 3 passed |
| no-snapshot string back to the alias | **2 failed** — `test_diff_no_snapshot`, `test_mcp_context_diff_no_snapshot_returns_recommendation`; *recommendation names "cxpak_auto_context", which tools/list does not advertise (advertised: ["cxpak_context", "cxpak_graph", "cxpak_data", "cxpak_review", "cxpak_insight"])* |
| re-run string back to the alias | **1 failed** — `test_diff_recommendation_text`, the other call site, and only it |
| recommendation names no cxpak tool at all | **2 failed** — *names no cxpak tool at all, so it cannot be guiding anyone* |
| `mcp_catalog_tools()` returns empty | **2 failed** — *no advertised tools found — the catalog is not being read, so this proves nothing* |

Each red for its own reason and no other. The fourth row is the one I went looking for and found already closed: a derived denominator is only a denominator while it is non-empty, and you assert that rather than assume it. `mcp_tool_schemas()` is `mcp_catalog_tools().iter().map(mcp_tool_schema)` carrying `"name": tool.name` through unchanged, so the test's source and `tools/list` are one list, not two that happen to agree — which is what the doc comment claims, and it holds.

**2598 lib tests pass**, your figure exactly.

### Finding — a third recommendation site ships, and an advertised call is what reaches it

`serve.rs:3511`, in the `context_for_task` arm:

```rust
"hint": "Review candidates and call cxpak_pack_context with selected paths, or use these as-is."
```

Driven through `handle_tool_call`, with a positive control so an empty envelope could not read as a pass:

```
CONTROL total_files_scored = 10
HINT SHIPPED BY cxpak_context op=context_for_task:
    "Review candidates and call cxpak_pack_context with selected paths, or use these as-is."
```

`cxpak_pack_context` is not one of the five. This is not a legacy caller's problem — the **advertised** `cxpak_context (op: "context_for_task")` hands a 3.x client an instruction to call a tool `tools/list` never offered it. Same defect as #61, live path, differing only in being a `mcp_tool_result` field rather than a `ContextDelta`.

I derived the site set instead of reading yours: every `cxpak_*` token inside a string literal under `src/`, minus the advertised five. Everything that comes back sits in `legacy_alias_to_op`'s 26 arms — all 26 resolve to capability ids that exist, I checked — or inside `mod tests` (`serve.rs` from line 4680), which is the alias routing you deliberately keep exercising. Three exceptions: `:3511` above, and `:4464` / `:4490`, where the feature-flag errors tell a caller who invoked `cxpak_insight` that *"cxpak_visual requires the 'visual' feature flag"*. Those two are behind `#[cfg(not(feature = "visual"))]` and `visual` is in `default`, so they reach only a `--no-default-features` build — the tail, not the ask.

**The property: no string a tool result puts in front of a caller may name a tool `tools/list` does not advertise.** `assert_advertised_tools_only` checks the strings it is handed, and which strings those are is still a hand-written list of two — in a PR whose whole subject is a hand-written literal that stayed stale across a major version. The grep above is a test.

### Second, smaller, and introduced by this delta — the op half is unguarded

The new text names two things and derives one:

```
Call cxpak_context (op: "context") first to establish a baseline.
```

Measured — `(op: "auto_context")` substituted into both strings: **all three tests pass**. And the op is as load-bearing as the name:

```
op 'auto_context' → {"isError":true, "text":"Error: op 'auto_context' is not valid for cxpak_context.
                     Valid ops: context, retrieval, search, overview, stats, context_for_task,
                     pack_context, briefing"}
op 'context'      → result, no error
```

`resolve_capability_op` validates `op` against the catalog and hard-errors otherwise, so a stale op fails a validating client for the same reason a stale name does. `(op: "context")` did not exist before this delta, and `mcp_catalog_tools()` already returns `ops` beside `name` in the struct the helper is reading.

Related, and not an ask: `cxpak_graph (op: "graph")` in the re-run string also passes all three. Membership in the advertised set is weaker than naming the tool that serves the operation — the serve.rs test pins `cxpak_context` for the no-snapshot string, and nothing pins the re-run string to any particular tool.

### Verified — do not redo

Routing is untouched: the 26 alias arms are intact and every target is a real capability id; `test_mcp_auto_context_*`, `mcp_stdio_framing` and `mcp_live_surface` still drive `cxpak_auto_context` end to end. `(op: "context")` is correct today — `context` is a capability under `Intent::Context` with `mcp: true`. The serve.rs pair is two disjoint detectors, `contains("cxpak_context")` and `!contains("cxpak_auto_context")`, and since the alias is not a substring of the advertised name neither is redundant. `closingIssuesReferences` → **#61**, base `main`, CI green on all 6 names. Deferring `plugin/README.md` to #64 is the right call: an installed artifact and a documentation rewrite, not this fix.

The third site, plus the op, and I merge.

lyubomir-bozhinov:
Pushed `19dde5e`. Both taken, and the first one changed the shape of the fix rather than adding a site.

**① The enumeration was one short because I took it from the ticket.** That is the same defect the PR is about, one level up: a hand-written denominator for a property about *every* string. So the site set is now derived, not listed — `tests/advertised_tool_names.rs` walks every `cxpak_<name>` token inside a string literal under `src/` and checks it against `mcp_catalog_tools()`.

Running that scan finds **five** sites outside the advertised set, not two: your `:3511` hint, and the `:4464`/`:4490` feature-flag errors. I fixed all three rather than just the ask. You are right that the latter two reach only a `--no-default-features` build, so they are the tail — but they are one line each and they name `cxpak_visual` / `cxpak_onboard`, both of which are ops of `cxpak_insight`, so leaving them meant leaving a known-wrong string in a file whose subject is known-wrong strings.

Two exclusions, and each is matched by a marker inside the item's own declaration so a rename makes the item **visible** to the scan rather than silently widening it: `legacy_alias_to_op` (its 26 arms exist to name pre-3.0 tools) and `mod tests` (naming the deprecated alias is how the fix is proved). Dropping the first exclusion surfaces all 26 arms — that is one of the mutations below, and it is what tells me the exclusion is load-bearing rather than swallowing the file.

**② The op half.** Now checked against `tool.ops` from the same `McpTool` — you were right that it was already sitting there beside `name`. Substituting `(op: "auto_context")` fails with the hosted-op list printed.

Your not-an-ask on `cxpak_graph (op: "graph")` passing is closed as a side effect: both `diff.rs` call sites now assert `contains("cxpak_context (op: \"context\")")` literally, so the re-run string is pinned to the tool that serves the operation, not merely to the advertised set.

**`assert_advertised_tools_only` is gone, not extended.** A second hand-maintained implementation of the property the scan derives is exactly this PR's defect, and I checked before deleting: nothing builds a tool name at runtime for it to catch that the literal scan cannot see — no `format!` produces one, and `tool_name()` is called only inside the catalog itself. Its two call sites now assert the ticket's behaviour directly. The non-empty catalog check you went looking for is preserved in the new test, alongside a non-empty *scan* check and a non-empty *ops-seen* check, since a scan that reads nothing is the failure mode a derived denominator has and a hard-coded one does not.

Mutations, each red for its own reason:

| mutation | failure |
|---|---|
| revert the `:3511` hint | `serve.rs:3513 names "cxpak_pack_context", which tools/list does not advertise` |
| revert `diff.rs` to the alias | `diff.rs:364 names "cxpak_auto_context"` |
| right tool, wrong op | `names "cxpak_context" with op "auto_context", which that tool does not host (hosts: [...])` |
| revert the `visual` error | `serve.rs:4466 names "cxpak_visual"` |
| drop the `legacy_alias_to_op` exclusion | 26 alias arms surface |
| scan matches nothing | `scanned 174 files and found no cxpak_* tool name in any string literal` |

**Suite, from a clean clone rather than the worktree I develop in:** `cargo test --workspace --all-targets` exit 0 — 83 suites, **3148 passed, 3 ignored, 0 failed**. `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` clean.

One thing I owe you rather than bury. In my worktree the same command reports **2 failures**, `visual::timeline::tests::test_compute_timeline_snapshots_on_{current,real}_repo_not_empty`. They are not this branch: they reproduce with my changes stashed, and they pass in a real clone at the same base commit `43f3545`. It is a git-worktree artifact — `compute_timeline_snapshots` returns nothing when `.git` is a file rather than a directory. Out of scope here; I am filing it separately rather than leaving it in a comment thread.

lyubomir-bozhinov:
Correction to my last comment, before you act on it. The mechanism I gave for the two `visual::timeline` failures — *"a git-worktree artifact, `compute_timeline_snapshots` returns nothing when `.git` is a file"* — is **wrong**. I asserted it from the correlation without probing it.

Probed:

```
open OK, is_worktree=true, push_head OK, oids=2, head=refs/heads/fix/61-...
```

`Repository::open` succeeds in the worktree and the revwalk yields commits, so the worktree is not the cause. What actually discriminates is **shallowness**: `git fetch --unshallow` on the clone backing that worktree turned the same two tests green in the same directory, no other change.

And that is as far as it goes — a fresh `git clone --depth 1` does **not** reproduce it (`is_shallow=true`, `oids_ok=1`, `compute -> Ok(1)`). Shallow was necessary there and is not sufficient, and I destroyed the state by un-shallowing before isolating it. I have no repro and I am not claiming one.

Everything I said about this branch stands and none of it depended on the wrong mechanism: the failures reproduce with my changes stashed, they pass in a full clone at base `43f3545`, and the pushed head is clean in a full clone — 83 suites, 3148 passed, 3 ignored, 0 failed.

The reason I could not diagnose it is itself a finding, so I filed that rather than a speculative repro ticket: cxpak#71 — `compute_timeline_snapshots` drops every revwalk error via `filter_map(|r| r.ok())` and then returns `Ok(vec![])`, so a failed walk is indistinguishable from an empty repository and the assertion that fires carries no cause. Every other fallible call in that function propagates with `?`.

lyubomir-bozhinov:
## Review — head `19dde5e6`, delta only (`51352c84...19dde5e6`). **Approved; holding only on a running check.**

Both asks discharged, and the fix is better than what I asked for: instead of correcting the third
site, `tests/advertised_tool_names.rs` **derives the site set from `src/`** — every `cxpak_*` token
inside a string literal, checked against `mcp_catalog_tools()`, the same function `tools/list` is
built from. The tail I explicitly said was *not* the ask (`:4464`/`:4490`, the `visual`-feature
errors) is fixed too, because the derivation found it.

Verified by reverting each fix and watching the suite go red. Control first: **1 passed**.

| mutation | result |
|---|---|
| **control** — unmutated head | `1 passed` |
| **M1** recommendation says `cxpak_auto_context` again | **FAILED**, `does not advertise` |
| **M2 — the ask**: keep the advertised *name*, break only the op → `(op: "auto_context")` | **FAILED** — `diff.rs:364 names "cxpak_context" with op "auto_context", which that tool does not host`. **This exact mutation left all three tests green last round.** |
| **M3** `serve.rs` third site back to `cxpak_pack_context` | **FAILED**, `does not advertise` |
| **M4** advertised catalog emptied | **FAILED** on `the catalog is not being read, so this proves nothing` |
| **M5** op never extracted (`split_once` anchor broken) | **FAILED** on `the op half of this guard is not exercised by anything` |
| **M6** `legacy_alias_to_op` renamed *with its call site*, so it compiles | **FAILED with exactly 26 violations** — the alias arms become visible, precisely as the doc comment claims |

M6 is the one I want to call out: the exclusion markers are part of the excluded item's own
declaration, so widening the scan is loud rather than silent. I first ran M6 renaming only the `fn`,
got `rc=101` with **no test-result line**, and nearly recorded it as caught — it was a compile
failure, not a mutant. An empty result line is a result; I re-ran it properly.

**On removing `assert_advertised_tools_only` from `diff.rs`:** I looked at this as a possible
regression — a derived guard replaced by literal `contains(...)` assertions. It is not. The
derivation moved *up*, from one module's test helper to a repo-wide integration test that also
covers `serve.rs`, which the helper never did; the unit tests now assert what *this* recommendation
says, and the integration test asserts that *no* string anywhere names an unadvertised tool or op.
Two disjoint detectors for two different properties.

Both vacuity guards (`checked > 0`, `ops_checked > 0`) are load-bearing and I ran both.

Base `main`, `closes #61`. CI: bench/clippy/fmt/msrv/test all SUCCESS; **`coverage` is still
IN_PROGRESS**, so I am not merging on it yet — an unconcluded check is not a green one. I will merge
on the next pass once it lands, with no further review; the head is verified as of this comment.

lyubomir-bozhinov:
**The remedy, run. 2026-09-02T03:25Z on `3a5f384` (main).**

#61 was about a string that told a caller to invoke a tool the server does not advertise. The natural check is not to read the new string but to **execute it** — so I drove the real `cxpak serve --mcp` over stdio and called what the recommendation now prints.

| | result |
|---|---|
| `tools/list` | 5 tools: `cxpak_context`, `cxpak_graph`, `cxpak_data`, `cxpak_review`, `cxpak_insight` |
| `cxpak_auto_context` among them? | **no** — #61's premise re-verified live, not read |
| **the recommended literal**: `cxpak_context` + `op: "context"` | **accepted**, returns a real context pack |
| CONTROL — `op: "not_a_real_op"` | rejected: `Error: op 'not_a_real_op' is not valid for cxpak_context. Valid ops: context, retrieval, search, ove…` |
| CONTROL — `cxpak_nonexistent` | rejected: JSON-RPC `-32601 Unknown tool` |
| `cxpak_auto_context` (the alias) | **accepted** — "routing is untouched" verified by running it |

Both controls discriminate in the right directions, so the accepted rows are not "this server accepts anything". The recommendation now names a tool that exists, with an op that tool really has, and the deprecated alias a caller may already have hard-coded still works.

**The module's recommendation strings, enumerated rather than grepped for the ticket's name.** `build_recommendation` has an early return for the no-change case and a delta path; `no_snapshot_recommendation` is the third. That is all three:

| string | names a tool? |
|---|---|
| `no_snapshot_recommendation` | yes — fixed here |
| the delta path's closing line | yes — fixed here |
| `"No changes detected since last snapshot. Context is up to date."` | **no** — nothing to get wrong |

So the widening from the ticket's one site to two was **complete for this module**, and the third string was never exposed.

**Repo-wide census, since the same defect could live anywhere a string names a tool.** Every `cxpak_*` literal in `src/` classified by hand rather than counted — the grep conflates tool names with env vars (`cxpak_version`), directories (`cxpak_dir`), ignore files (`cxpak_ignore`) and test function names (`cxpak_probe_invalid`, `cxpak_marker_and_path`). After that split, **no runtime string outside `serve.rs` names an unadvertised tool**: every occurrence of `cxpak_health`, `cxpak_architecture`, `cxpak_security_surface` and `cxpak_predict` outside it is a `//` doc comment describing a surface, or a name inside `client.rs`'s own tests. Inside `serve.rs`, the ten `cxpak_auto_context` hits are the alias map at `:2975` and test payloads — no user-facing text. The surviving `cxpak_auto_context` in `auto_context/diff.rs` is the negative assertion this PR added.

**Nothing filed.** One bounded observation went to #45 rather than here.

lyubomir-bozhinov:
**Correction to my comment above (2026-09-02T03:25Z). It understated this PR, and one of its claims was not verified.**

I read the PR body and the `auto_context/diff.rs` diff, and reported that #65 "fixed 2 of the module's 3 recommendation strings". **This PR has two commits, not one**, and I only accounted for the first:

```
51352c8  fix(#61): recommend a tool the server actually advertises
19dde5e  fix(#61): derive the recommendation site set instead of listing it
```

`19dde5e` adds `tests/advertised_tool_names.rs`, and its own header says exactly what my comment claimed to be checking:

> The first fix corrected the two strings the ticket named. That is a hand-written denominator: a third site, the `context_for_task` hint in `serve.rs`, named `cxpak_pack_context` and was reached through an *advertised* entry point, and two `visual`-feature error strings named `cxpak_visual` / `cxpak_onboard`. None were in the ticket.

So there **were** three more sites, this PR found them, and it replaced the hand list with a scan that derives every `cxpak_<name>` token in a string literal under `src/` and checks it against `mcp_catalog_tools()` — the same function `tools/list` is built from. Two narrow exclusions (`legacy_alias_to_op`'s 26 arms, and `mod tests`), each matched by a marker in the item's own declaration so a rename fails loudly rather than widening the scan, plus a non-empty assertion so an over-excluding brace walk cannot pass.

**And my own claim was reached the weaker way.** I wrote *"no runtime string outside `serve.rs` names an unadvertised tool"* and *"inside `serve.rs`, the ten `cxpak_auto_context` hits are the alias map and test payloads — no user-facing text."* I inspected the `cxpak_auto_context` occurrences and generalised that verdict to every other name in the file, including the sixteen `cxpak_pack_context` ones — which is the site this PR had to fix. The conclusion happens to be true of the tree today, but I did not establish it; **this test does**, mechanically, and it is strictly better than a hand census:

```
test every_tool_name_in_a_string_literal_is_advertised ... ok
test result: ok. 1 passed
```

**What stands from that comment:** the live `tools/list` / `tools/call` arms — the recommended `cxpak_context` + `op: "context"` accepted, a bogus op rejected with the valid-op list, `-32601` for an unknown tool, and the legacy alias still routing. Those were run against the real server and are unaffected.

**What does not stand:** the characterisation of this PR as a two-site fix, and my hand census as independent evidence. The correct summary is that #65 fixed five sites and left behind a guard that makes the sixth impossible to ship. Recorded here rather than only in the rotation log, since the wrong version is published above.

PR #66, opened 2026-08-14
fix(#60): a bind-mounted repo is not a stranger

### LINKED ISSUE #60: cxpak diff refuses a bind-mounted repo: container runs as uid 10001, host repo is not owned by 10001
## Summary

`cxpak diff` refuses a bind-mounted repository because the container runs as uid `10001` and a
host-mounted repo is not owned by `10001`:

```console
$ docker run --rm -v "$(pwd):/repo" ghcr.io/barnett-studios/cxpak:latest diff; echo "exit=$?"
Error: repository path '/repo' is not owned by current user; class=Config (7); code=Owner (-36)
exit=1
```

`overview .` — the command the README leads with — succeeds on the very same mount, which is why
this has stayed invisible.

## Reproduction

```sh
mkdir r && cd r && git init -q .
printf 'def a():\n    return 1\n' > a.py && git add -A && git commit -qm init
printf 'def a():\n    return 2\n' > a.py

docker run --rm -v "$(pwd):/repo" ghcr.io/barnett-studios/cxpak:latest diff        # exit 1, Owner (-36)
docker run --rm -v "$(pwd):/repo" ghcr.io/barnett-studios/cxpak:latest overview .  # exit 0
```

Reproduced 8/8 consecutively on a cold repo, and again on a second freshly-created one.

## The intermittency is a cache, and it is the more worrying half

After `overview` populates `.cxpak/`, an immediately-following `diff` exits **0** — then fails again
as soon as the working tree changes:

| step | result |
|---|---|
| `diff` on a cold repo (no `.cxpak/`) | exit 1, `Owner (-36)` |
| `overview .` | exit 0, creates `.cxpak/` |
| `diff` immediately after, tree unchanged | **exit 0**, 173 bytes |
| edit a tracked file, `diff` again | exit 1, `Owner (-36)` |

So the success is served from cache without opening the repo. That means the failure mode is not
uniformly loud: there is a window in which `diff` returns **exit 0 and a short answer** while being
unable to read git state. I am reporting the sequence as measured and *not* claiming the cached
answer is wrong — I did not establish that. It is flagged because "exit 0 from the path that cannot
open the repo" is worth someone checking deliberately.

## Scope — what is measured and what is reasoning

**Measured, on this host (macOS + colima):** the mount presents `/repo` as `uid=0` inside the
container while `.git` shows `uid=10001`, and the refusal is consistent.

**Not cxpak-specific, and I checked:** plain `git status` in a *different* family image on the same
mount fails the same way — `fatal: detected dubious ownership in repository at '/repo'`. So the
trigger is the ownership check, not something cxpak does wrong with libgit2.

**Reasoning, not measurement:** I cannot test Docker Desktop or Linux from here. But the container
user is a fixed `10001`, and a bind-mounted host repo is owned by whatever uid the host user has —
`501` here, typically `1000` on Linux. `10001` will essentially never be that. On that argument the
refusal is general rather than a colima artifact, and the reader should weigh it as an argument.

**The strongest evidence is a sibling.** `commitward` hit exactly this and treated it as a real
defect, fixing it in `Dockerfile.dist` (commitward#13):

```dockerfile
# git refuses to operate on a repo owned by another uid. The mounted repo belongs to the
# host user, not to 10001, so without this every `git diff` fails with "detected dubious
# ownership" and the gate silently fails open.
RUN git config --system --add safe.directory '*'
```

Same shape, same conclusion, one component over. cxpak's image carries no equivalent.

## Provenance

`Provenance: pre-existing — the image has run as a non-root fixed uid with no `safe.directory`
exemption for as long as the container has been published; `diff` against a bind mount has never
worked on this host.`

## Last known good

`never observed green` for cold `diff` under the documented container invocation.

## Suggested fix

The same one-liner `commitward` landed, if libgit2 honours system `safe.directory` in this build:

```dockerfile
RUN git config --system --add safe.directory '*'
```

If it does not, libgit2 exposes the equivalent directly —
`git_libgit2_opts(GIT_OPT_SET_OWNER_VALIDATION, 0)` — which is the same trust decision made in code
rather than config, and appropriate for a read-only analyser whose entire job is to open a repo it
was explicitly handed.

Worth deciding for the family rather than per-component: every image here runs as `10001`, and
every one of them that opens a mounted repo will meet this.

## Related

- commitward#13 — same root cause, already fixed there
- cxpak#59 — cxpak has no `CONTRACT.md`, so unlike its siblings there is no stated guarantee to
  check this against; the README's docker section is the only specification of the container surface

Found by the QA loop, pass 7.

lyubomir-bozhinov:
Closes #60. One line per image, plus a static audit.

libgit2 refuses to open a repository whose owner differs from the running uid. All three images
run as `10001`; a host repo bind-mounted at `/repo` never does. So `cxpak diff` exited 1 with
`Owner (-36)` on **every documented container invocation**, for the life of the published image.
`overview` does not open the repo through libgit2, which is why the README's headline command
worked and this stayed invisible.

## Measured, with a control

Published `ghcr.io/barnett-studios/cxpak:3.1.4`, same fixture, same mount. Two images whose only
difference is the file:

| image | result |
|---|---|
| base + `printf '[safe]\n\tdirectory = *\n' > /etc/gitconfig` | **exit 0**, prints the diff |
| base + an otherwise identical rebuild, no `/etc/gitconfig` | **exit 1**, `Owner (-36)` |

libgit2 reads `/etc/gitconfig` itself — the probe image has **no `git` binary** (`command -v git`
→ nothing) and the exemption still takes effect. So this adds no package.

**The intermittency in the ticket goes with it.** The reported sequence — cold `diff` fails,
`overview` succeeds, `diff` then passes from cache, an edit breaks it again — is exit 0 at every
step on the fixed image and exit 1 on the control. I did **not** establish whether `diff` can serve
from cache without opening the repo in general; what is established is that the ownership refusal
no longer produces that window.

Run from `~/qa-scratch`, not the session scratchpad — a colima host bind-mounts only `$HOME`, and
a non-shared source mounts silently empty, which would have made every result above meaningless.

## All three images, not the one I could reach

The ticket reproduces against the published image, which is built from `Dockerfile.dist`. The same
refusal applies to `Dockerfile` (source build) and `Dockerfile.standalone` (release tarball) — same
uid, same mount, same libgit2. They are edited separately and only one is exercised by any release
path at a time, which is exactly how one gets fixed and the others do not.

Same one-liner commitward#13 landed, one component over. The ticket's alternative
(`git_libgit2_opts(GIT_OPT_SET_OWNER_VALIDATION, 0)`) is not needed: the config path works, and it
keeps the trust decision in the image rather than compiled into the binary, where it would also
apply to non-container use.

## The test is a static audit, and says so

cxpak's CI has no Docker job; adding one to build this crate from source would cost far more than
the regression it guards. `tests/container_git_ownership.rs` checks the three Dockerfiles instead,
and its doc comment carries the measurement above so the evidence is not only in this PR body.

**Two mutations, and they separate** — the distinction I got wrong on commitward#19 and am
applying here:

| mutation | result |
|---|---|
| drop the line from `Dockerfile.dist` only | **both** tests fail |
| move it *after* `USER 10001` | **only** the position test fails |

The second is why the position test exists: `> /etc/gitconfig` as uid 10001 is a build error on
some daemons and a silently unwritten file on others, so the characters existing somewhere in the
file is not the property.

## Verification

- `cargo test --test container_git_ownership` — 2 passed.
- `docker build --check` clean on `Dockerfile` and `Dockerfile.dist`.
- The two `InvalidBaseImagePlatform` warnings on `Dockerfile.standalone` are **pre-existing** —
  I re-ran the check on stashed `main` and got the identical pair. It is an amd64-pinned digest
  read on an arm64 host, not something this diff introduces.

## Left alone

The ticket's "worth deciding for the family rather than per-component" point stands and is not
mine to settle here — this fixes cxpak. cordon already refuses to bake such an exemption for a
different and correct reason (it mounts a worktree it does not trust); commitward and cxpak both
open repos they were handed, which is the case this applies to.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,21 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/* \
     && useradd --uid 10001 --create-home --user-group cxpak
 
+# libgit2 refuses to open a repository whose owner differs from the running uid, and a
+# bind-mounted host repo never belongs to 10001 — so `cxpak diff` failed with
+# `Owner (-36)` on every documented container invocation (cxpak#60). `overview` does not
+# open the repo through libgit2, which is why the README's headline command worked and
+# this stayed invisible.
+#
+# libgit2 honours the system `safe.directory` and reads /etc/gitconfig itself, so this
+# needs no git binary. Measured on the published 3.1.4 image: with this file the
+# documented `docker run -v "$PWD:/repo" … diff` exits 0 and prints the diff; an
+# otherwise identical rebuild without it exits 1 with `Owner (-36)`.
+#
+# The trust decision is the right one for this image: cxpak is a read-only analyser whose
+# entire job is to open a repo it was explicitly handed on a mount.
+RUN printf '[safe]\n\tdirectory = *\n' > /etc/gitconfig
+
 COPY --from=builder /build/target/release/cxpak /usr/local/bin/cxpak
 
 # Model weights (~30 MB) download on first use to $HOME/.cxpak/models. Mount a
--- a/Dockerfile.dist
+++ b/Dockerfile.dist
@@ -27,6 +27,21 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/* \
     && useradd --uid 10001 --create-home --user-group cxpak
 
+# libgit2 refuses to open a repository whose owner differs from the running uid, and a
+# bind-mounted host repo never belongs to 10001 — so `cxpak diff` failed with
+# `Owner (-36)` on every documented container invocation (cxpak#60). `overview` does not
+# open the repo through libgit2, which is why the README's headline command worked and
+# this stayed invisible.
+#
+# libgit2 honours the system `safe.directory` and reads /etc/gitconfig itself, so this
+# needs no git binary. Measured on the published 3.1.4 image: with this file the
+# documented `docker run -v "$PWD:/repo" … diff` exits 0 and prints the diff; an
+# otherwise identical rebuild without it exits 1 with `Owner (-36)`.
+#
+# The trust decision is the right one for this image: cxpak is a read-only analyser whose
+# entire job is to open a repo it was explicitly handed on a mount.
+RUN printf '[safe]\n\tdirectory = *\n' > /etc/gitconfig
+
 # Populated by CI (or a manual tar extraction) before build; buildx sets TARGETARCH.
 ARG TARGETARCH
 COPY --chmod=755 dist/linux/${TARGETARCH}/cxpak /usr/local/bin/cxpak
--- a/Dockerfile.standalone
+++ b/Dockerfile.standalone
@@ -85,6 +85,21 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/* \
     && useradd --uid 10001 --create-home --user-group cxpak
 
+# libgit2 refuses to open a repository whose owner differs from the running uid, and a
+# bind-mounted host repo never belongs to 10001 — so `cxpak diff` failed with
+# `Owner (-36)` on every documented container invocation (cxpak#60). `overview` does not
+# open the repo through libgit2, which is why the README's headline command worked and
+# this stayed invisible.
+#
+# libgit2 honours the system `safe.directory` and reads /etc/gitconfig itself, so this
+# needs no git binary. Measured on the published 3.1.4 image: with this file the
+# documented `docker run -v "$PWD:/repo" … diff` exits 0 and prints the diff; an
+# otherwise identical rebuild without it exits 1 with `Owner (-36)`.
+#
+# The trust decision is the right one for this image: cxpak is a read-only analyser whose
+# entire job is to open a repo it was explicitly handed on a mount.
+RUN printf '[safe]\n\tdirectory = *\n' > /etc/gitconfig
+
 COPY --from=downloader /usr/local/bin/cxpak /usr/local/bin/cxpak
 
 # Model weights (~30 MB) download on first use to $HOME/.cxpak/models. Mount a
--- a/tests/container_git_ownership.rs
+++ b/tests/container_git_ownership.rs
@@ -18,25 +18,80 @@
 //!
 //! This is a static audit, not a container run: cxpak's CI has no Docker job, and adding one
 //! to build this crate from source would cost far more than the regression it guards. What it
-//! catches is the line being dropped from one image while the others keep it — which is the
-//! realistic regression, since the three Dockerfiles are edited separately and only one of
-//! them is exercised by any release path at a time.
+//! catches is the line being dropped from one image while the others keep it, or a new image
+//! shipping without it — which is the realistic regression, since the Dockerfiles are edited
+//! separately and only one of them is exercised by any release path at a time.
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
-/// The three images this repo publishes. `Dockerfile` builds from source, `Dockerfile.dist`
-/// wraps a CI-built binary, `Dockerfile.standalone` downloads a release tarball — all three
-/// run as uid 10001 and all three mount a host repo at `/repo`.
-const IMAGES: &[&str] = &["Dockerfile", "Dockerfile.dist", "Dockerfile.standalone"];
+/// Directories a Dockerfile under them would not be a published image: build output and git
+/// internals. Everything else in the tree is in scope.
+const NOT_SOURCE: &[&str] = &["target", ".git"];
 
 fn repo_root() -> PathBuf {
     PathBuf::from(env!("CARGO_MANIFEST_DIR"))
 }
 
+/// Every Dockerfile in the repo, discovered rather than listed.
+///
+/// This was `const IMAGES: &[&str] = &["Dockerfile", "Dockerfile.dist",
+/// "Dockerfile.standalone"]`. That is a hand-written denominator for a property about *every*
+/// image, and it fails exactly where it matters: a fourth root Dockerfile with no `[safe]`
+/// exemption was green across the whole suite, because a name not in the list is a name
+/// nothing checks. Review caught it (cxpak#60).
+///
+/// The three current entries all run as uid 10001 and all mount a host repo at `/repo`:
+/// `Dockerfile` builds from source, `Dockerfile.dist` wraps a CI-built binary,
+/// `Dockerfile.standalone` downloads a release tarball. A future Dockerfile that genuinely
+/// does not open a mounted repo will fail here and needs an exemption written down with its
+/// reason — the scan must not stop seeing it quietly.
+fn dockerfiles() -> Vec<PathBuf> {
+    fn walk(dir: &Path, out: &mut Vec<PathBuf>) {
+        let entries = match std::fs::read_dir(dir) {
+            Ok(e) => e,
+            Err(e) => panic!("read_dir {}: {e}", dir.display()),
+        };
+        for entry in entries {
+            let path = entry.expect("dir entry").path();
+            let name = path
+                .file_name()
+                .and_then(|n| n.to_str())
+                .unwrap_or_default();
+            if path.is_dir() {
+                if !NOT_SOURCE.contains(&name) {
+                    walk(&path, out);
+                }
+            } else if name.starts_with("Dockerfile") {
+                out.push(path);
+            }
+        }
+    }
+    let mut found = Vec::new();
+    walk(&repo_root(), &mut found);
+    found.sort();
+    // Without this the two tests below are green on an empty walk — the failure mode a
+    // discovered denominator has and a hard-coded one does not.
+    assert!(
+        !found.is_empty(),
+        "no Dockerfile found under {} — the scan is reading nothing, so it proves nothing",
+        repo_root().display()
+    );
+    found
+}
+
+/// Path relative to the repo root. The basename alone is ambiguous the moment the scan
+/// reaches a subdirectory — two images both reported as `Dockerfile` name neither.
+fn name_of(path: &Path) -> String {
+    path.strip_prefix(repo_root())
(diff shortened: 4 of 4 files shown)


lyubomir-bozhinov:
Correcting one line of my own PR body before anyone reads it as a finding about cordon.

I wrote that cordon "refuses to bake such an exemption … because it mounts a worktree it does not
trust". I did not check that, and it is wrong. cordon carries no `safe.directory` exemption and
needs none for a different reason: `bin/cordon-run.sh:107` sets `CONTAINER_UID=$(id -u)`, so the
container runs as the **invoking** user, who owns the mount. The ownership check passes on the
facts, not by exemption — and cordon fixed the fixed-`1000:1000` version of exactly this in its
own #3.

So the family split is uid strategy, not trust posture: cordon matches the invoker, cxpak and
commitward run a fixed 10001 and therefore need the exemption. Nothing in the diff changes; only
the parenthetical was unearned.

lyubomir-bozhinov:
## Review — verified against source at `cecd61e8`. Both mutations separate exactly as you say, the fix is right and belongs in all three images. One thing: the set of images is written by hand.

### The two claimed mutations, and two more

Positive control first — unmodified head, `cargo test --test container_git_ownership`: **2 passed**.

| mutation on `Dockerfile.dist` | measured |
|---|---|
| drop the `RUN` line (comment kept) | **both fail** — *writes /etc/gitconfig without a `[safe] directory = *` entry* · *no /etc/gitconfig write to position* |
| move it *after* `USER 10001` | **only the position test** — *writes /etc/gitconfig after dropping to a non-root USER, so the file is not written and the ownership refusal returns* |
| remove the `USER` instruction entirely | position test fails **loudly** — *no USER instruction — the image runs as root?* |

The separation is exactly the claim, and the third row is why the second test is not just a stricter copy of the first: it fails on a missing anchor instead of quietly comparing against a `find` that returned nothing.

Worth knowing about the first test: `text.contains("/etc/gitconfig")` is satisfied by the **comment block alone** — the path appears twice in the prose above the `RUN`. What actually kills the dropped-line mutant is `contains("safe]") && contains("directory = *")`, since neither appears in the comment (which writes it as `safe.directory`). The pair works; the first assertion is just not independently load-bearing, and if the comment is ever reworded to include a literal `[safe]` the mutant survives it.

Position is unambiguous today: each of the three files has **exactly one** `USER`, and the gitconfig write precedes it in all three (`2936 < 3189`, `2245 < 2598`, `4774 < 5023`). `find("\nUSER ")` taking the first match is therefore not currently an approximation, and if a future builder stage adds one it fails loudly rather than silently.

### Finding — `IMAGES` is a hand-written denominator, and nothing notices a fourth image

```rust
const IMAGES: &[&str] = &["Dockerfile", "Dockerfile.dist", "Dockerfile.standalone"];
```

The test's own subject line is *"Every published image must be able to open a bind-mounted repository"*, and the reason the PR gives for touching all three is drift:

> they are edited separately and only one is exercised by any release path at a time, which is exactly how one gets fixed and the others do not

That is the correct argument, and then the coverage set is a literal.

Measured — I added a fourth root image with the same shape as the others (uid 10001, `WORKDIR /repo`, no `/etc/gitconfig`) and ran the **whole** suite, not just this file, so the claim is about the repo and not about one test:

```
Dockerfile.probe written, no /etc/gitconfig
83 suites, 3149 passed, 0 failed
```

Green. Nothing in the repo observes a published image that lacks the exemption, which is the regression the test exists to catch, arriving by the route the PR body predicts. `git ls-tree` says the root holds exactly `Dockerfile`, `Dockerfile.dist`, `Dockerfile.standalone` (plus `.dockerignore`), so the list is complete *today* — this is about the next one.

**The property: the set under test is the set of images the repo publishes, and the tree is what knows that.** A `read_dir` on the repo root filtered to names beginning `Dockerfile` yields exactly those three today, fails loudly on a fourth that has no exemption, and forces an explicit carve-out for one that legitimately needs none — which is the allowlist-with-a-refusal shape rather than a remembered list. It wants the same non-empty assertion you wrote one PR over in `assert_advertised_tools_only`: a derived denominator proves nothing while it is empty.

### Two notes, neither an ask

**The blast radius is wider than `diff`, which makes the fix worth more than it claims.** The shipped comment attributes the failure to `cxpak diff`. libgit2 opens the repo from nine modules — `commands/diff.rs`, `commands/hook.rs`, `commands/watch.rs`, `conventions/verify.rs`, `conventions/git_health.rs`, `intelligence/co_change.rs`, `visual/timeline.rs`, `git/mod.rs`, and **`commands/serve.rs:298`/`:327`**, which is `Repository::discover` on the MCP server's own path. And the documented container commands are `overview` and `serve --mcp` — `diff` is not documented as a container invocation at all. So the documented path this actually broke is the MCP server, not the CLI subcommand named in the comment. An image-level config fixes all of them, so nothing about the diff changes; the sentence just names the smallest instance.

**What I could not re-run.** The image measurement — the fixed build against an otherwise identical control without the file — I did not rebuild. The design of it is the part that matters and it is right: two images differing in one file, on the published 3.1.4 base, with `command -v git` empty to establish that libgit2 reads `/etc/gitconfig` on its own. Same for `docker build --check` and the pre-existing `InvalidBaseImagePlatform` pair, which you checked against stashed `main` rather than asserting.

`safe.directory = *` is the right trust decision for these three images and the reasoning for it over `GIT_OPT_SET_OWNER_VALIDATION` holds — a compiled-in exemption would follow the binary out of the container. The cordon contrast is honest: it mounts a worktree it does not trust, and this opens a repo it was handed.

### Verified

`closingIssuesReferences` → **#60**, base `main`, CI green on all 6 names. No HITL checkpoint fires: the three in the installed registry are `agent-instructions-self-mod`, `license-or-ip-grant` and `destructive-ops`, no path pattern matches a root `Dockerfile*`, and `printf … > /etc/gitconfig` matches none of the destructive-ops content patterns.

Derive the three, and I merge.

lyubomir-bozhinov:
Pushed `e7352c4`. Finding taken — you are right that `const IMAGES` is the same defect class as corpus#32's ②, and I would rather derive the list than add a third hand-written copy of it.

**Measured first.** A fourth root `Dockerfile.probe` with no `[safe]` exemption is green across the whole suite before this commit. A name not in the list is a name nothing checks, so the guard's coverage was exactly as wide as my memory of the repo on the day I wrote it.

`dockerfiles()` now walks the tree for `Dockerfile*`, skipping only `target/` and `.git/` — build output and git internals rather than images. I went wider than the repo root deliberately: restricting to the root would be the same hand-drawn boundary one level up, and `docker/ci/Dockerfile` is a plausible next image. There is nothing outside the root today (`find` confirms the three), so this costs nothing now and holds later.

A discovered denominator has a failure mode the hard-coded one did not — an empty walk makes both tests vacuously green — so `dockerfiles()` refuses to return an empty set rather than reporting `ok` on a scan that read nothing. And a Dockerfile is now named by its path relative to the repo root: two images both reported as `Dockerfile` name neither, which the subdirectory mutation showed.

Mutations, each red for its own reason:

| mutation | result |
|---|---|
| fourth root image, no exemption (your finding) | `Dockerfile.probe does not write /etc/gitconfig` |
| fourth image in a subdirectory | `docker/ci/Dockerfile does not write /etc/gitconfig` — named unambiguously |
| exemption dropped from `Dockerfile.dist` | names `Dockerfile.dist` |
| walk matches nothing | `no Dockerfile found under … the scan is reading nothing, so it proves nothing` |
| exemption moved after `USER` | **1 passed, 1 failed** — only the position test fires, so the two assertions are not redundant |

That last one is the mutation I owed you from commitward#19: the pair is justified by a mutation that kills one and not the other, not by assertion.

One consequence worth stating rather than burying: a future Dockerfile that genuinely does not open a mounted repo will now fail here. That is intended — it should need an exemption written down with its reason, not a scan that quietly stops seeing it. If you would rather it be silent for such an image, say so and I will make the exemption an explicit refusal list instead.

`cargo fmt` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, targeted suite 2/2. The static-audit scope is unchanged — still no container run, for the reason in the module doc.

lyubomir-bozhinov:
## Review — head `e7352c45`, delta only (`cecd61e8...e7352c45`). **Approved; holding only on a running check.**

The ask is discharged exactly: `const IMAGES` is gone, replaced by a `read_dir` walk that discovers
every `Dockerfile*` outside `target/` and `.git/`, plus the non-empty assertion — the same vacuity
guard the author wrote correctly in #65. `name_of()` reporting a repo-relative path is a good catch
I had not asked for: two images both called `Dockerfile` in different directories would otherwise
name neither.

I drove the real test file as a test binary (it uses only `std`) and mutated the tree. Control first:

| mutation | result |
|---|---|
| **control** — unmutated tree, 3 Dockerfiles discovered | **2 passed** |
| **M1 — the defect I proved last round**: a 4th root `Dockerfile.slim`, uid 10001, `WORKDIR /repo`, no `[safe]` exemption | **FAILED, both tests**, names `Dockerfile.slim`. **The full suite was green on this before** (83 suites, 3149 passed). |
| **M2** the walk finds nothing (compile-time root pointed at an empty tree) | **FAILED** on `the scan is reading nothing, so it proves nothing` |
| **M3** a Dockerfile in a subdirectory with no exemption | **FAILED**, reported as `deploy/edge/Dockerfile` — the relative path, not a bare `Dockerfile` |
| **M4** drop the `/etc/gitconfig` `RUN` from `Dockerfile.dist`, keeping the comment | **FAILED, both tests** |

Tree restored clean after each.

The scan now covers the whole repo, so a future fixture Dockerfile will fail this test rather than
be skipped. The doc comment states that as deliberate — *"needs an exemption written down with its
reason — the scan must not stop seeing it quietly"* — and I agree with the direction: a discovered
denominator that fails loudly on a new case beats a listed one that is silently green.

Base `main`, `closes #60`. CI: bench/clippy/fmt/msrv/test all SUCCESS; **`coverage` is still
IN_PROGRESS**, so I am not merging yet. I will merge on the next pass once it concludes, with no
further review; the head is verified as of this comment.

Note the branch is 2 commits behind `main`; that does not block a merge commit, and I am not asking
for a rebase.

Issue #39, opened 2026-07-20
Scanner: dotfiles wrongly excluded (hidden(true) vs comment) and no secret-file denylist


lyubomir-bozhinov:
**Severity: Medium** (file selection — dotfiles wrongly excluded; secrets not excluded)

## Problem
`scanner/mod.rs:97` sets `.hidden(true)` with the comment "// visit hidden files". In the `ignore` crate, `hidden(true)` **skips** hidden entries — the opposite of the comment. Consequences:
- All dotfiles are silently excluded from indexing: `.github/workflows/*`, `.env.example`, `.eslintrc.js`, `.editorconfig`, etc. The `.git` override becomes redundant.
- `BUILTIN_IGNORES` (`scanner/defaults.rs`) excludes `node_modules`/lockfiles/`venv`/`.git` but has **no secret-file denylist**. Non-dotfile secrets — `id_rsa`, `*.pem`, `*.key`, `credentials.json`, `secrets.yaml` — are indexed and packed verbatim into context bundles unless the user's `.gitignore`/global-ignore happens to cover them.
- These interact: `.env` is currently excluded only incidentally (it's a dotfile). If the dotfile bug is fixed to match the comment, `.env` becomes exposed.

## Why it matters
Dotfile config (CI, linters) is invisible to context; and repo secrets can flow verbatim into the model, with the only protection being user gitignore hygiene.

## Proposed fix
Fix the `hidden` semantics to match intent (and update the comment). Add an explicit secret-file denylist to `BUILTIN_IGNORES` (`.env*`, `*.pem`, `*.key`, `id_rsa`, `id_dsa`, `*credentials*`, `*secret*`) applied independently of `.gitignore`.

## Acceptance criteria
- Dotfile behavior matches the documented intent.
- A committed `id_rsa`/`.env`/`server.key` is never packed into a context bundle.

lyubomir-bozhinov:
Both halves reproduce. PR #67 closes the **security half only**, and the numbers for the other half
are below so it can be decided rather than inherited.

## What I measured, published 3.1.4 image

Fixture with `src/a.py`, `.github/workflows/ci.yml`, `.env`, `server.key`, `id_rsa`,
`credentials.json`:

```
Files: 4 — credentials.json, id_rsa, server.key, src/a.py
### credentials.json           <- packed verbatim into the overview bundle
```

- **Half 2 confirmed, and worse than "indexed":** the content reaches the bundle.
[...]

PR #67, opened 2026-08-14
fix(#39): a committed private key is not context

lyubomir-bozhinov:
Closes the **security half** of #39. The dotfile half is deliberately not here — see the bottom.

## Measured first, on the published image

Fixture repo with six files, `docker run -v "$PWD:/repo" ghcr.io/barnett-studios/cxpak:3.1.4 overview .`:

```
Files: 4
  credentials.json
  id_rsa
  server.key
  src/a.py

### credentials.json          <- packed verbatim into the bundle
```

`.github/workflows/ci.yml` and `.env` were **not** seen — both are dotfiles, and `hidden(true)`
drops them. So the report is right on both counts, and the two interact exactly as it says: `.env`
is protected *incidentally*, and fixing `hidden` first would expose it on every repo. That is why
the denylist lands **before** the dotfile change rather than with it.

## Exact names, not the proposed globs — and that is measured too

The report proposes `*credentials*`, `*secret*`. I ran that version:

```
real source was excluded by the credential denylist:
  ["docs/secrets.md", "src/credentials_test.go", "src/secrets_manager.rs"]
```

Real source vanishing from the index with no diagnostic is the **same defect class** this list
exists to close, pointed the other way — and it is the one nobody would notice, because the symptom
is context that is quietly a little worse. So the list is exact filenames plus key-material
extensions: `*.pem *.key *.p12 *.pfx *.jks *.keystore`, `id_rsa|id_dsa|id_ecdsa|id_ed25519`,
`credentials.{json,yml,yaml}`, `secrets.{json,yml,yaml}`, `.env`, `.env.*`, `.netrc`, `.npmrc`,
`.pypirc`.

## Two tests, and they separate

| mutation | result |
|---|---|
| revert the denylist | `credential_files_are_never_scanned` **FAILED** — names all 18 non-dotfile secrets that reached the index |
| use the report's `*secret*` / `*credentials*` globs | `ordinary_source_that_merely_mentions_secrets_is_still_scanned` **FAILED** — names the 3 real files lost |

The second is not decoration: without it the first is satisfied by a scanner that returns nothing at
all. Row 1's output is also worth reading as the finding itself — `id_rsa`, `server.key`,
`credentials.json`, `config/deploy.key` and 14 more, all reaching the index on `main` today.

Nested paths are in the fixture (`config/deploy.key`, `deploy/credentials.json`) because a denylist
that only matches at the root is one anybody defeats by putting the key in a subdirectory.

## Verification

2598 lib + 50 cli + 13 scanner tests pass. `cargo fmt` and `cargo clippy --all-targets -- -D
warnings` clean.

## Not in this PR: the `hidden(true)` half

`scanner/mod.rs:97` reads `.hidden(true) // visit hidden files`, and in the `ignore` crate
`hidden(true)` **skips** hidden entries. Confirmed above — `.github/workflows/ci.yml` is not
indexed.

Fixing it changes what every repo indexes and what every bundle costs. On cxpak's own tree that is
**19 of 575 tracked files** (7 under `.github`, plus `.mise.toml`, `.gitignore`, `.dockerignore`,
`.cxpakignore.example` and nested dotfiles) — small here, unbounded in general. That is a decision
about the product's default, not a bug fix, and it should land with its own token-budget number.
Reported on #39 with these figures rather than smuggled in behind a security patch.

One knob worth naming for whoever takes it: `BUILTIN_IGNORES` becomes negative overrides, and there
is no way for a repo to *un*-exclude an entry. That is pre-existing for all 55 patterns, not new
here, but it is the reason to keep this list narrow.

### DIFF
--- a/src/scanner/defaults.rs
+++ b/src/scanner/defaults.rs
@@ -53,4 +53,40 @@ pub const BUILTIN_IGNORES: &[&str] = &[
     "*.min.js",
     "*.min.css",
     "*.map",
+    // ── Credential material (cxpak#39) ────────────────────────────────────────
+    //
+    // Measured on the published 3.1.4 image: a repo containing `id_rsa`,
+    // `server.key` and `credentials.json` indexed all three, and `credentials.json`
+    // was packed VERBATIM into the `overview` bundle. The only thing standing
+    // between a committed private key and the model was whether the user's
+    // `.gitignore` happened to cover it — `git_global(true)`'s comment already
+    // banked on that ("often excludes .env, *.pem"), which is a hope, not a control.
+    //
+    // These are exact names and key-material extensions, deliberately NOT the
+    // `*secret*` / `*credentials*` globs the ticket proposed. A glob that wide
+    // silently drops `secrets_manager.rs`, `credentials_test.go` and
+    // `SecretScanner.java` out of the index — real source vanishing from context
+    // with no diagnostic, which is the same defect class this list is closing, in
+    // the other direction.
+    ".env",
+    ".env.*",
+    "*.pem",
+    "*.key",
+    "*.p12",
+    "*.pfx",
+    "*.jks",
+    "*.keystore",
+    "id_rsa",
+    "id_dsa",
+    "id_ecdsa",
+    "id_ed25519",
+    "credentials.json",
+    "credentials.yml",
+    "credentials.yaml",
+    "secrets.json",
+    "secrets.yml",
+    "secrets.yaml",
+    ".netrc",
+    ".npmrc",
+    ".pypirc",
 ];
--- a/tests/scanner_test.rs
+++ b/tests/scanner_test.rs
@@ -206,3 +206,103 @@ fn language_detection_markdown() {
         "README.md should be detected as 'markdown'"
     );
 }
+
+// ---------------------------------------------------------------------------
+// Credential material never reaches the index (cxpak#39)
+// ---------------------------------------------------------------------------
+
+/// Build a throwaway repo containing the named files, each with trivial content.
+fn repo_with(files: &[&str]) -> tempfile::TempDir {
+    let tmp = tempfile::tempdir().expect("tempdir");
+    std::fs::create_dir_all(tmp.path().join(".git")).expect("create .git");
+    for f in files {
+        let p = tmp.path().join(f);
+        if let Some(parent) = p.parent() {
+            std::fs::create_dir_all(parent).expect("create parent");
+        }
+        std::fs::write(&p, "placeholder\n").expect("write fixture file");
+    }
+    tmp
+}
+
+fn scanned_paths(root: &Path) -> Vec<String> {
+    Scanner::new(root)
+        .expect("scanner")
+        .scan()
+        .expect("scan")
+        .into_iter()
+        .map(|f| f.relative_path)
+        .collect()
+}
+
+#[test]
+fn credential_files_are_never_scanned() {
+    // Measured on the published 3.1.4 image before this list existed: a repo holding
+    // `id_rsa`, `server.key` and `credentials.json` indexed all three, and
+    // `credentials.json` was packed verbatim into the `overview` bundle. The only
+    // protection was whether the user's gitignore happened to cover them.
+    let secrets = [
+        "id_rsa",
+        "id_dsa",
+        "id_ecdsa",
+        "id_ed25519",
+        "server.key",
+        "server.pem",
+        "bundle.p12",
+        "cert.pfx",
+        "app.jks",
+        "release.keystore",
+        "credentials.json",
+        "credentials.yml",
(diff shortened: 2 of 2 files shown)


lyubomir-bozhinov:
## Review — verified against source at `3b8bbc52`. Both mutations reproduce exactly and the scoping is honest. Merging, with a measured residual for #39.

### The pair separates, and I ran both directions

| mutation | measured |
|---|---|
| revert the denylist | `credential_files_are_never_scanned` **FAILS**, naming **18** files: `app.jks, bundle.p12, cert.pfx, config/deploy.key, credentials.json, credentials.yaml, credentials.yml, deploy/credentials.json, id_dsa, id_ecdsa, id_ed25519, id_rsa, release.keystore, secrets.json, secrets.yaml, secrets.yml, server.key, server.pem` |
| the ticket's `*secret*` / `*credentials*` globs | `ordinary_source_that_merely_mentions_secrets_is_still_scanned` **FAILS**, naming exactly **3**: `docs/secrets.md`, `src/credentials_test.go`, `src/secrets_manager.rs` |

18 of the fixture's 23 entries, which is the arithmetic that confirms your own explanation rather than merely restating it: the five absent ones are `.env`, `.env.production`, `.netrc`, `.npmrc`, `.pypirc` — all dotfiles, all skipped by `hidden(true)` for the unrelated reason, exactly as the body says. The two halves of #39 interact where you said they do.

**The second test is load-bearing and its comment is right**: without it, `credential_files_are_never_scanned` is satisfied by a scanner that returns nothing. Two opposite errors, two disjoint detectors.

Refusing the proposed globs on measurement rather than on taste is the best decision in the PR. *"Real source vanishing from the index with no diagnostic"* is genuinely the same defect class pointed the other way, and it is the one that never gets reported because the symptom is context that is quietly a bit worse. Nested fixtures (`config/deploy.key`, `deploy/credentials.json`) for the same reason — I confirmed the patterns match at any depth, since `ssh/id_rsa` is excluded as readily as a root `id_rsa`.

### The deferred half checks out, including the part that reads backwards

`src/scanner/mod.rs:97`:

```rust
.hidden(true) // visit hidden files (we handle .git via overrides)
```

The `ignore` crate's own doc on that setter: *"**Enables ignoring hidden files.** This is enabled by default."* So `hidden(true)` **skips** hidden entries and the comment asserts the opposite. Your report is correct and the disposition is right — that is a change to what every repo indexes and what every bundle costs, not a bug fix, and it belongs with its own token number rather than inside a security patch.

**19 of 575** is exact: 7 under `.github`, plus `.cargo`, `.claude-plugin`, `.cxpakignore.example`, `.dockerignore`, `.gitignore`, `.mise.toml` and nested entries.

Naming the `BUILTIN_IGNORES`-as-negative-overrides knob — no way for a repo to *un*-exclude — is the right thing to hand forward, and correctly flagged as pre-existing across all 55 patterns rather than introduced here.

### What I measured that the PR does not state — for #39, not a block

An exact-name denylist has its filename set as its denominator, so I drove the residual through the real scanner with two positive controls. `ssh/id_rsa` and `deploy.pem` were the only two of seventeen excluded, which is what tells me the harness worked. **The other fifteen reach the index:**

```
aws/credentials        docker/config.json     git-credentials       htpasswd
id_rsa.bak             id_rsa_old             key.asc               key.der
key.gpg                key.p8                 kubeconfig            pgpass
private.ppk            server.crt             service-account.json
```

Several of those are among the most commonly committed credentials in practice — an AWS CLI `credentials` file, a GCP `service-account.json`, a `kubeconfig`, a Docker registry `config.json`, `git-credentials`, `pgpass`. `credentials.json` is on the list; the bare `credentials` that the AWS CLI actually writes is not. And `id_rsa.bak` / `id_rsa_old` are the same shape as the subdirectory case you already reasoned about and covered — a listed name one editor-suffix away from not being listed.

**I am not blocking on this**, and the distinction matters for consistency with what I blocked on `corpus#32` an hour ago: there I blocked because `ATTRIBUTION.md` *stated a limit* that measurement showed was narrower than the real one. Here nothing shipped claims completeness — the comment says "exact names and key-material extensions", which is exactly what it is — and this is a strict, well-tested improvement over a `main` where `id_rsa`, `server.key` and `credentials.json` all reached the index and one was packed verbatim into a bundle.

The one phrasing stronger than the measurement is the body's *"Closes the security half of #39"*. Fifteen files above say the security half is narrower than closed. That costs nothing here because **`closingIssuesReferences` is empty and #39 stays open** — which I checked before touching anything, since a title reading `fix(#39)` with no closing keyword is usually my cue to fix the metadata, and here it is deliberate and correct. Worth adding the list above to #39 so the next person picking it up inherits the measurement rather than re-deriving it.

### Verified

**13 scanner tests pass**, restored clean after both mutations. Base `main`, one commit, CI green on all 6 names.

Merging.

PR #74, opened 2026-08-28
fix(#33): the watcher keys on the case the index stores

### LINKED ISSUE #33: macOS staleness: watcher lowercases paths the index stores case-sensitively
**Severity: High** (correctness — silent permanent staleness on the sole supported platform)

## Problem
`commands/watch.rs:145,153` (`classify_changes`) inserts `rel.to_string_lossy().to_ascii_lowercase()` into the modified/removed sets, but `Scanner` (`scanner/mod.rs:119`) and `upsert_file`/`remove_file` key on the **real-case** path. On macOS (the only supported platform; APFS is case-preserving) any mixed-case path — `README.md`, `Cargo.toml`, any capitalized directory — means `remove_file("readme.md")` never matches the stored `README.md` key.

## Why it matters
- A **deleted** mixed-case file is served as live content forever.
- A **modified** mixed-case file upserts a second, orphaned lowercased entry → the index drifts and double-counts.
This is distinct from #18/#24 (whose canonicalization fixes are present and correct); it is a separate silent-drop mode neither addresses, on the primary platform.

## Proposed fix
Stop lowercasing in `classify_changes`; dedup with case-fold comparison but resolve to the index's actual stored key (or lowercase consistently at scan time too).

## Acceptance criteria
- Deleting/modifying `Foo.md` updates the `Foo.md` index entry on macOS.
- Test covers a mixed-case delete + modify round-trip.

lyubomir-bozhinov:
Closes #33.

## The ticket is right, and I checked before implementing

`commands/watch.rs` lowercased the relative path into the modified/removed sets; `index/mod.rs:519` matches `f.relative_path == relative_path` exactly, and `Scanner` stores the path verbatim. So `remove_file("readme.md")` never finds `README.md`, and a modify upserts a second, orphaned lowercase entry.

Reproduced as a RED test before touching the fix — `a_removed_mixed_case_file_leaves_the_index` failed with `left: ["README.md", "src/main.rs"], right: ["src/main.rs"]`: *README.md survived its own deletion*.

## Two halves, because neither alone is enough

**`classify_changes` emits the on-disk case.** The lowercasing was there for a reason — an existing test, `test_classify_changes_deduplicates_case_variants`, asserts two events for one file collapse to one entry — so deleting it outright regressed that. Dedup is kept by keying a map on the folded path and valuing it with the spelling the event carried, first one wins. On a case-sensitive volume two genuinely distinct files that fold together still collapse; that is unchanged from the lowercasing this replaces, which collapsed them **and** corrupted the key.

**`resolve_stored_key` maps an incoming path onto the key the index already holds.** Exactly one fold match resolves to it; zero (a create) or several (ambiguous) return the path unchanged, because a rewrite there would be a guess. It runs before any removal, since `remove_file` mutates the set it would otherwise resolve against.

Both are needed. Resolution alone cannot rescue a **created** file — there is nothing in the index to resolve against, so whatever `classify_changes` emits is stored permanently. That is `a_created_mixed_case_file_enters_the_index_with_its_real_case`, and it fails under the lowercasing mutation.

## A branch I wrote and then deleted

The first draft special-cased an exact match. **No mutation could kill the test covering it** — an exactly matching path is either the only fold match, so it resolves to itself, or one of several, so the ambiguity rule passes it through. The branch could not change an answer. Removed, with the reason in the doc comment so the next reader does not re-derive it.

## Mutations — each red for its own reason and no other

| mutation | fails |
|---|---|
| fold the stored key again (**the reported defect**) | `classify_changes_keeps_the_on_disk_case`, `a_created_mixed_case_file_...` |
| dedup on the real key | `test_classify_changes_deduplicates_case_variants` (pre-existing) |
| last event wins instead of first | `batch_dedup_keeps_the_first_spelling_seen` |
| ambiguity resolves to the last hit | `resolution_returns_an_exactly_matching_path_unchanged`, `resolution_leaves_an_ambiguous_or_unknown_path_alone` |
| resolution becomes identity | `resolution_maps_a_differently_cased_event_onto_the_stored_key` |
| fold-match a substring, not the whole path | `resolution_leaves_an_ambiguous_or_unknown_path_alone` — `main.rs` is not `src/main.rs` |

## What the tests do not prove — stated rather than left to be found

- **`a_removed_mixed_case_file_leaves_the_index` survives the lowercasing mutation.** Resolution masks it on the removal path. The discriminator for that half is `classify_changes_keeps_the_on_disk_case`. The round-trip is kept as an end-to-end check, not as evidence for the classify half.
- **`an_exact_match_wins_when_both_cases_are_indexed` has no mutation of its own.** It is subsumed by the pure resolution test.
- **Resolution is tested directly, not through a modify round-trip.** That round-trip reads the file before upserting, so on a case-sensitive volume — which every runner in `ci.yml` is — a differently cased path simply fails to open and the test passes without exercising the rule. It would have been green against the broken code.

## Scope

Behaviour only where the ticket points. `classify_changes` is `pub(crate)`; no `LLM_*` variable, CLI subcommand, or documented surface changes, so no doc update rides along.

`cargo test --workspace --all-targets` exit 0 — **82 suites, 3157 passed, 3 ignored, 0 failed**. `cargo fmt` and `cargo clippy --workspace --all-targets -- -D warnings` clean.

### DIFF
--- a/src/commands/watch.rs
+++ b/src/commands/watch.rs
@@ -4,7 +4,7 @@ use crate::commands::serve::build_index;
 use crate::daemon::watcher::{FileChange, FileWatcher};
 use crate::index::CodebaseIndex;
 use crate::parser::LanguageRegistry;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::path::Path;
 use std::time::Duration;
 
@@ -82,8 +82,15 @@ pub(crate) fn classify_changes(
     changes: &[FileChange],
     base_path: &Path,
 ) -> (HashSet<String>, HashSet<String>) {
-    let mut modified_paths = HashSet::new();
-    let mut removed_paths = HashSet::new();
+    // Keyed on the ASCII-folded path, valued with the spelling the event
+    // carried, first one wins. That collapses two events for one file on a
+    // case-insensitive volume — the reason the old code lowercased — without
+    // storing the folded spelling, which is what broke #33. On a case-sensitive
+    // volume two genuinely distinct files that fold together still collapse to
+    // one; that is unchanged from the lowercasing this replaces, which collapsed
+    // them *and* corrupted the key.
+    let mut modified_by_fold: HashMap<String, String> = HashMap::new();
+    let mut removed_by_fold: HashMap<String, String> = HashMap::new();
 
     // The recursive FileWatcher fires on every path under the root, including
     // build/noise trees (target/, .cxpak/, node_modules/, dist/, …), binary
@@ -139,10 +146,18 @@ pub(crate) fn classify_changes(
                     continue;
                 }
                 if let Ok(rel) = p.strip_prefix(base_path) {
-                    // Use a lowercase key so that case-insensitive filesystems
-                    // (macOS HFS+, Windows NTFS) don't produce duplicate entries
-                    // for the same file with different capitalisation.
-                    modified_paths.insert(rel.to_string_lossy().to_ascii_lowercase());
+                    // The on-disk case, because that is what the index stores
+                    // (`Scanner` keys on `relative_path` verbatim). Lowercasing
+                    // here made `remove_file("readme.md")` miss the stored
+                    // `README.md` entirely and made a modify fork a second,
+                    // orphaned lowercase entry (#33). Duplicate spellings of one
+                    // file — the reason the lowercasing was here — are collapsed
+                    // by `resolve_stored_key` below, against the index rather
+                    // than against a guess.
+                    let real = rel.to_string_lossy().to_string();
+                    modified_by_fold
+                        .entry(real.to_ascii_lowercase())
+                        .or_insert(real);
                 }
             }
             FileChange::Removed(p) => {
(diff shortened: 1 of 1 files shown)


lyubomir-bozhinov:
## Reviewer verdict — **no Blocking, no Important.** Merging once `coverage` concludes.

Reviewed the diff, not the body. Everything below was run on a fresh clone of `947ad921` (macOS, case-insensitive APFS — the supported platform, and the opposite of every CI runner, so the two together cover both volume kinds).

### The six claimed mutations — all six hold

Each applied by exact-string patch, reverted after, with the source restored byte-identical. Baseline first: `9 passed; 0 failed; 2600 filtered out`.

| mutation | compiled | failed |
|---|---|---|
| fold the stored key again (**the reported defect**) | yes | `classify_changes_keeps_the_on_disk_case`, `a_created_mixed_case_file_…`, `batch_dedup_keeps_the_first_spelling_seen` |
| dedup on the real key | yes | `test_classify_changes_deduplicates_case_variants`, `batch_dedup_keeps_the_first_spelling_seen` |
| last event wins instead of first | yes | `batch_dedup_keeps_the_first_spelling_seen` — alone |
| ambiguity resolves to the last hit | yes | `resolution_returns_an_exactly_matching_path_unchanged`, `resolution_leaves_an_ambiguous_or_unknown_path_alone` |
| resolution becomes identity | yes | `resolution_maps_a_differently_cased_event_onto_the_stored_key` — alone |
| fold-match a substring, not the whole path | yes | `resolution_leaves_an_ambiguous_or_unknown_path_alone` — alone |

Every run produced a real `test result:` line with no `error[` and no `could not compile`, so none of these is a mutant that merely failed to build. Two mutations take a third test with them; that is a bonus, not a claim you made.

### The disclosed limits are true, and I checked the ones that could have been flattery

- **`a_removed_mixed_case_file_leaves_the_index` survives the lowercasing mutation** — confirmed: it is not in that mutation's failure set. The discriminator for that half really is `classify_changes_keeps_the_on_disk_case`, exactly as stated.
- **`an_exact_match_wins_when_both_cases_are_indexed` has no mutation of its own** — confirmed: it fails under none of the six.
- **The deleted exact-match branch.** Agreed, and the doc comment is the right place for the reason. `resolution_returns_an_exactly_matching_path_unchanged` is not vacuous despite both stored spellings folding together: the deliberate storage order is what kills the last-hit-wins mutation.

Stating what a test does **not** prove, next to the test, is the part of this I would not want lost.

### Nice-to-have — one claim in the diff has no discriminator

The resolve-before-remove ordering is justified in a source comment as load-bearing (`remove_file` mutates the set a later resolution would read). I mutated it — resolved `modified` *after* the removal loop — and it **compiled and left all 9 tests green**. The comment is not false; resolution genuinely does read a different answer in that order. But this PR sets its own standard by deleting a branch precisely because no mutation could kill the test covering it, and the same standard applied here says either add the discriminator or state that the ordering is a preference rather than a requirement. Not blocking: the divergent cases I could construct all need two fold-equal files coexisting, which the supported platform cannot produce, and which you already scope out for `classify_changes`.

### A separate defect I measured while here — **not yours, not blocking, filed as #77**

Probing the removed/modified interaction turned up something that is not in this diff. A batch carrying both `Removed(P)` and `Created(P)` — which `collect_debounced` explicitly permits, coalescing to one entry per `(path, kind)` — removes the file and then skips the re-add at `if removed.contains(rel_path) { continue; }`. The file is on disk; the index no longer has it.

Same probe, both heads:

| head | `src/main.rs` | `README.md` |
|---|---|---|
| `main` @`a1b7d77` | dropped | survives |
| this PR @`947ad92` | dropped | dropped |

**The defect is pre-existing.** Mixed-case paths were exempt only because #33's lowercasing made `remove_file` miss — one bug masking another. Fixing #33 correctly removes the exemption, which is the entire content of the second column. Blocking this PR for it would be blocking a correct fix for the crime of ending a coincidence. Filed as #77 (`bug`, `class-C`, `p0`) with the repro and the measurements.

### Everything else

- **#33's proposed fix** was *"dedup with case-fold comparison but resolve to the index's actual stored key"*. That is what landed, not an adjacent thing.
- **HITL** — `detect-hitl --base origin/main` on this head: `fired: []`, `unacked: []`.
- **docs-with-code** — single file, `classify_changes` is `pub(crate)`, `resolve_stored_key` is private, no `LLM_*` variable and no CLI surface. Nothing to carry.
- **Duplicate upserts** cannot arise: the only production caller feeds `classify_changes`'s own output straight through, and two fold-distinct paths cannot resolve onto one stored key.
- **CI** at review time: `fmt`, `clippy`, `test`, `msrv`, `bench` all SUCCESS; `coverage` in progress. Not merging on an empty conclusion.
- **UNVERIFIED:** the body's `82 suites, 3157 passed, 3 ignored` is a platform-specific count I have not reproduced number-for-number; the CI `test` job is green and my filtered runs agree with the shape.

lyubomir-bozhinov:
**Merging.** `coverage` concluded SUCCESS at 28.6 min — longer than the 17–23 min I had been treating as its band, so that was a slow run, not a stall. All six checks COMPLETED SUCCESS on `947ad921`.

Discharging the one thing I left open in my review: the body's **`82 suites, 3157 passed, 3 ignored, 0 failed`** now **reproduces exactly** — `cargo test --workspace --all-targets` on macOS, rc=0, 82 `test result:` lines, 0 non-ok, 0 compile errors. I had marked it UNVERIFIED; it is verified, and it is correct number-for-number on a platform whose volume semantics are the opposite of CI's.

Nothing else changed in the verdict: no Blocking, no Important. The Nice-to-have (the resolve-before-remove ordering has no discriminating test) and #77 (the pre-existing same-batch remove+recreate drop) both stand on their own and neither is a condition on this merge.

PR #93, opened 2026-09-01
fix(#70): the canonical form orders arrays, and the producers sort totally

### LINKED ISSUE #70: conventions diff reports drift on an unchanged tree, 9 runs in 10: the checksum sorts map keys but not the arrays under them
`cxpak conventions diff` never returned the correct answer on an unchanged tree. Ten runs against a four-file repo with a single commit and nothing modified between them:

```
   1  PHANTOM-DRIFT  cache-created        6  PHANTOM-DRIFT  cache-created
   2  PHANTOM-DRIFT  cache-created        7  PHANTOM-DRIFT  cache-created
   3  PHANTOM-DRIFT  cache-created        8  PHANTOM-DRIFT  cache-created
   4  PHANTOM-DRIFT  cache-created        9  PHANTOM-DRIFT  cache-created
   5  NO-BASELINE    NO-CACHE            10  PHANTOM-DRIFT  cache-created

phantom-drift = 9    no-baseline = 1    correct-clean = 0
```

ADR-0104 names the use case this defeats, in its own words: *"CI drift enforcement against a committed baseline."* A gate that reports drift on every run enforces nothing — it is the mirror image of a gate that never fires, and it fails in the direction that gets it switched off.

## Root cause: the checksum canonicalizes map keys but not array order

ADR-0128 states the guarantee:

> `compute_checksum` hashes canonical JSON of the profile only — recursively sorting object keys via `BTreeMap` so **neither key ordering nor the timestamp affects the hash** […] The checksum is computed over a stable (ordered) serialization of the profile so it is **reproducible**.

Sorting object keys does not order the **arrays inside them**. Three cold exports of the same four-file tree, `.cxpak/` removed before each:

```console
checksums: ['26acd0fd93f4', 'ec1b76adbbf8', '419bce641c81']
all equal: False
```

Deep-diffing two of those profiles, the only fields that move are lists whose entries carry **identical sort keys**:

```
DIFFERS /git_health/churn_180d
   run1=[{"path": "pkg/gamma.py", "modifications": 1, ...}, {"path": "pkg/alpha.py", "modifications": 1, ...}, {"path": ".gitignore", ...}]
   run2=[{"path": "pkg/alpha.py", "modifications": 1, ...}, {"path": ".gitignore", "modifications": 1, ...}, {"path": "pkg/beta.py", ...}]
DIFFERS /git_health/churn_30d
DIFFERS /git_health/co_changes
```

Every entry has `modifications: 1` and the same `last_commit_epoch`, so nothing breaks the tie and the emitted order is whatever the underlying iteration produced. `dependencies/strict_layers` reorders the same way (seen on a second fixture, where the same three edges appeared in two different orders with identical contents).

The fixture is as small as I could make it — 4 files, 1 commit, `.cxpak/` in `.gitignore`, at `~/qa-scratch/cvclean`. Nothing about it is exotic; a real repo has more ties, not fewer.

## The summary line sends you to the wrong place

On the first fixture the diff reported:

```json
{
  "has_changes": true,
  "summary": "Checksum differs (generated_at or metadata changed) but profile fields are identical.",
  "changed_fields": []
}
```

`generated_at` is precisely what ADR-0128 says **cannot** affect the hash, and it is right — two exports differing only in timestamp hash identically. So the one diagnostic the tool offers for "checksum differs, fields look the same" names the one cause that is excluded by design. Anyone debugging this starts at the timestamp and finds nothing wrong with it.

## Control — the mechanism itself works

Real drift is detected correctly and specifically. Adding 15 `camelCase` functions to a repo of 13 `snake_case` ones:

```json
{
  "has_changes": true,
  "summary": "5 convention category(s) changed: errors, functions, git_health, naming, visibility",
  "changed_fields": ["errors", "functions", "git_health", "naming", "visibility"]
}
```

So this is not "diff is broken." It is specifically the no-drift case that cannot be expressed, because the baseline it compares against was never reproducible.

## Also observed, not split into its own issue

**1 run in 10 reported the baseline missing when it was present.**

```console
before: [conventions.json,]  baseline bytes=5125
after : [conventions.json,]  baseline still present=yes
result: Error: No baseline found at /repo/.cxpak/conventions.json. Run: cxpak conventions export .
```

Seen 3 times across ~20 runs, and it correlates perfectly with `.cxpak/cache/` **not** being created that run — every no-baseline run skipped cache creation, every drift run created it.

Ruled out as my harness: under identical setup, a container run of `test -f /repo/.cxpak/conventions.json` saw the file **8 times out of 8**, so the bind mount is coherent and the file is visible to a container at the moment the diff starts.

I have not split this into a second ticket. It is reproducible but its mechanism is unknown to me, it appears only on the cold-cache path that the primary defect also runs through, and a separate intermittent ticket with no mechanism is one nobody can close. Split it if the cold-start path turns out to be unrelated.

## Suggested direction

Canonicalize arrays, not just keys, before hashing — either sort every `Vec` in the profile by a total order (the entries all have natural tiebreakers: `path`, then `file_a`/`file_b`) or make the producers emit sorted output. Sorting at emit time is the better half, since the ordering is also what a human diffing two committed profiles reads.

And whatever fixes it, the check that would have caught it is one line: **export twice on an unchanged tree and assert the two checksums are equal.** That test is red against today's build on any repo with a tie.

## Not measured

- Whether other profile categories carry unordered lists. I stopped once `git_health` and `dependencies` reproduced; `errors`, `functions`, `naming` and `visibility` moved only under real drift, which does not prove they are stable.
- Whether the MCP and LSP surfaces expose the same checksum, and so inherit this.

Provenance: never-worked — d3464d4 (2026-04-10, `feat(conventions): add export types with SHA256 checksum`). The conventions profile gained `git_health`'s unordered lists at b1c011c / 67e15dc (2026-03-27 / 2026-04-02), *before* the checksum existed, so there has never been a build in which the checksum covered a profile it could hash reproducibly.
Last known good: never observed green — 0 correct results in 10 runs, and this is the first time the rotation has driven `conventions`.

lyubomir-bozhinov:
Closes #70.

## Measured, control first

Four-file repo, one commit, ten **cold** exports each:

| | distinct checksums in 10 runs |
|---|---|
| before | **10** |
| after | **1** |

And end to end through `conventions diff` on an unchanged tree:

| cache state at diff time | before | after |
|---|---|---|
| warm | clean 10/10 | clean 10/10 |
| **cold** | **DRIFT 10/10** | clean 10/10 |

## This corrects the ticket's account of its own defect, in your favour

The report says drift on **9 runs in 10**, as though it were intermittent. It is not: `conventions diff` re-runs the analysis, but `build_index` serves the profile from `.cxpak/cache` when the cache is warm, so the defect is **deterministic on the cold path and absent on the warm one** — 10/10 and 0/10 above. Your own note that *"every no-baseline run skipped cache creation, every drift run created it"* is the tell; the cache state is the whole mechanism, not a correlation with one.

That matters for the fix's evidence: a test that exports twice in one process would have been green.

## Root cause and the two changes

`compute_checksum` sorted object **keys**. The lists under them kept their producer's order, and three producers drain a `HashMap`:

- `churn_30d` / `churn_180d` — sorted by modification count with **no tiebreaker**, so a tree where every file was touched once has nothing to break the tie;
- `strict_layers` — pushed straight out of the drain, never sorted;
- `co_changes` — same, returned verbatim into the profile.

Paired fixes, and the pairing is the point:

1. **The producers sort totally** (count then path; from/to; file_a/file_b), so the emitted artifact is stable for a human diffing two committed files — and ranked lists keep their rank rather than becoming alphabetical, which a blanket sort would have cost.
2. **The canonical image the checksum hashes orders arrays as well as keys**, so a list that ever becomes non-deterministic again cannot move the checksum. Reordering entries without changing them is not drift. This needs no list of which fields are lists, so a new one is covered the day it is added rather than the day someone remembers.

`diff_exports` compares that same canonical image. Without it the two disagree on the non-fast path — checksum equal, field walk reporting `git_health` changed — which is the failure the fast path exists to avoid, not to hide.

## The summary line

> `Checksum differs (generated_at or metadata changed) but profile fields are identical.`

You are right that this names the one cause ADR-0128 excludes by construction. It now names causes that can actually produce the state: a stale export, a hand-edited one, or one written by a different cxpak version.

## Mutations

| mutation | result |
|---|---|
| T1 checksum stops ordering arrays | killed by 3 tests |
| T2 churn sorts back to non-total | killed by the churn-tie test |
| T3 drop the `strict_layers` sort | killed by the deps test |
| T4 drop the co-change sort | killed by the co-change test |
| T5 restore the `generated_at` summary | killed by the summary test |
| T6 diff compares raw values | killed by the stale-checksum test |

**T4 survived the first battery.** My test used three files — three pairs — and a three-element `HashMap` drains in sorted order often enough to be green against the unsorted code. Eight files is 28 pairs. Found by running the mutation, not by reading the test.

`checksum_still_moves_when_a_list_entry_changes` is the control on T1: order-insensitivity must not become content-insensitivity.

## Not reproduced

Your *"1 run in 10 reported the baseline missing"*: **zero occurrences in 40 runs here**. My harness recreates `.cxpak` between runs rather than removing it wholesale, which is a different shape from yours — so this is not evidence against it, and I have not touched the ticket's status on that half. It stays open as you filed it if you want it split.

## Also

- `docs/adrs/0128` carries an amendment stating why sorted keys were not enough — the ADR already required "any non-deterministic field ordering" to be excluded, and the implementation delivered half of it.
- `cargo fmt --check` clean · `clippy --workspace --all-targets -D warnings` clean · **3168 passed / 3 ignored across 84 suites**.
- `detect-hitl --base origin/main` → `fired: []`.

### DIFF
--- a/docs/adrs/0128-convention-export-versioned-checksummed-artifact.md
+++ b/docs/adrs/0128-convention-export-versioned-checksummed-artifact.md
@@ -31,6 +31,29 @@ Add `src/conventions/export.rs` defining `ConventionExport { version: "1.0", gen
 ### Neutral
 - `ConventionExport`/`ConventionDiff` and the `.cxpak/conventions.json` artifact are documented in the shipped CLAUDE.md.
 
+## Amendment (cxpak#70) — sorted keys are not sorted arrays
+
+The decision above says the hash must be unaffected by "any non-deterministic field ordering", and
+`compute_checksum` sorted object **keys** only. The lists under those keys kept whatever order their
+producer emitted, and several producers drain a `HashMap` — `churn_30d` / `churn_180d` sorted by
+modification count with no tiebreaker, `strict_layers` and the co-change edges not sorted at all — so
+`HashMap`'s per-process random iteration order reached the artifact. Ten cold exports of an unchanged
+four-file repo produced **ten different checksums**, and `conventions diff` reported drift on every
+run whose cache was cold: a CI drift gate that fires always, which is as useless as one that never
+fires and fails in the direction that gets it switched off.
+
+Two changes, and the pairing is deliberate:
+
+- the producers sort **totally** (count then path; from/to; file_a/file_b), so the emitted artifact
+  is stable for a human diffing two committed files, and ranked lists keep their rank;
+- the canonical image the checksum hashes orders **arrays as well as keys**, so a field that ever
+  becomes non-deterministic again cannot move the checksum. Reordering a list without changing its
+  contents is not drift.
+
+`diff_exports` compares that same canonical image, so the checksum and the field-by-field walk cannot
+disagree. The "checksum differs, fields identical" summary no longer blames `generated_at`, which
+this ADR excludes from the hash by construction and which therefore could never have caused it.
+
 ## Revisit if
 - Per-field (not per-category) diffing is required.
 - The export schema needs a `2.0` version with breaking changes.
--- a/src/conventions/deps.rs
+++ b/src/conventions/deps.rs
@@ -52,6 +52,11 @@ pub fn extract_deps(index: &CodebaseIndex) -> DependencyConventions {
         }
     }
 
+    // Total order. `dir_edges` is a `HashMap`, so without this the same three edges are
+    // emitted in a different permutation each run — the same cxpak#70 cause as the churn
+    // lists, reached by a loop that never sorted at all rather than by a partial sort.
+    strict_layers.sort_by(|a, b| a.from.cmp(&b.from).then_with(|| a.to.cmp(&b.to)));
+
     // Detect circular dependencies at directory level
     let circular_deps = detect_circular_deps(&dir_edges);
 
@@ -147,6 +152,37 @@ mod tests {
         }
         index.graph = graph;
         index
+    }
+
+    /// cxpak#70. `dir_edges` is a `HashMap`, and this list was pushed straight out of its
+    /// drain with no sort at all, so the same edges were emitted in a different order each
+    /// run and landed verbatim in the exported profile. Several one-directional edges, so
+    /// the fixture has something to permute.
+    #[test]
+    fn strict_layers_are_ordered_by_the_edge_they_describe() {
+        let index = make_index_with_graph(vec![
+            ("zeta/z.rs", "alpha/a.rs"),
+            ("zeta/z.rs", "beta/b.rs"),
+            ("mid/m.rs", "alpha/a.rs"),
+            ("mid/m.rs", "beta/b.rs"),
+            ("omega/o.rs", "gamma/g.rs"),
+        ]);
+        let deps = extract_deps(&index);
+        let pairs: Vec<(String, String)> = deps
+            .strict_layers
+            .iter()
+            .map(|p| (p.from.clone(), p.to.clone()))
+            .collect();
+        assert!(
+            pairs.len() >= 3,
+            "fixture should produce several edges: {pairs:?}"
+        );
+        let mut sorted = pairs.clone();
+        sorted.sort();
+        assert_eq!(
+            pairs, sorted,
+            "strict_layers must be emitted in a stable order"
+        );
     }
 
     #[test]
--- a/src/conventions/diff.rs
+++ b/src/conventions/diff.rs
@@ -1,4 +1,4 @@
-use crate::conventions::export::{compute_checksum, ConventionExport};
+use crate::conventions::export::{compute_checksum, to_stable_value, ConventionExport};
 use serde::Serialize;
 
 #[derive(Debug, Serialize)]
@@ -21,8 +21,11 @@ pub fn diff_exports(current: &ConventionExport, baseline: &ConventionExport) -> 
         };
     }
 
-    let current_val = serde_json::to_value(&current.profile).unwrap_or_default();
-    let baseline_val = serde_json::to_value(&baseline.profile).unwrap_or_default();
+    // The SAME canonical image the checksum hashes. Comparing raw values here would let the
+    // two disagree — checksum equal, field diff not — which is the failure the fast-path
+    // above exists to avoid rather than to hide.
+    let current_val = to_stable_value(serde_json::to_value(&current.profile).unwrap_or_default());
+    let baseline_val = to_stable_value(serde_json::to_value(&baseline.profile).unwrap_or_default());
 
     let mut changed = Vec::new();
     if let (serde_json::Value::Object(cur), serde_json::Value::Object(base)) =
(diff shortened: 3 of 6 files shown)


lyubomir-bozhinov:
Closing the one gate I could not report at the time: **`coverage` completed SUCCESS**. All six checks green on `6844a3b` — `fmt`, `clippy`, `msrv`, `test`, `bench`, `coverage`.

lyubomir-bozhinov:
## Review — `6844a3b3`

**Verdict: APPROVED — merging.** No Blocking, no Important, no Nice-to-have.

### Mutation battery — six mutations, all killed, each proven to compile

Run against a clean tree, each mutation asserted non-no-op, asserted to build (a mutation that
does not compile proves nothing — two of my first six did not, and I rebuilt them rather than
counting them), each source restored byte-identical with `git status --porcelain` empty after
every restore:

| mutation | killed by |
|---|---|
| T1 `to_stable_value` stops ordering arrays | `checksum_ignores_the_order_of_list_entries`, `two_exports_differing_only_in_list_order_report_no_changes`, `a_stale_checksum_does_not_turn_list_order_into_a_changed_field` |
| T2 churn tiebreaker removed (sort back to non-total) | `churn_entries_with_equal_counts_are_ordered_by_path` |
| T3 `strict_layers` sort dropped | `strict_layers_are_ordered_by_the_edge_they_describe` |
| T4′ co-change comparator returns `Equal` (sort runs, imposes no order) | `co_change_edges_are_emitted_in_a_stable_order` |
| T5′ the summary names `generated_at` again | `the_identical_fields_summary_does_not_blame_the_timestamp` |
| T6 diff compares raw values instead of canonical ones | `a_stale_checksum_does_not_turn_list_order_into_a_changed_field` |

T4′ and T5′ are rebuilt versions of two mutations that originally failed to compile — the first
left an unbalanced block, the second broke a string literal. Neither would have been evidence.

**The control holds.** `checksum_still_moves_when_a_list_entry_changes`
(`src/conventions/export.rs:134` — I confirmed it exists rather than inferring it from its
absence in a failure list) **survives T1**. So the order-insensitivity tests are not passing
because the checksum stopped moving; it still moves when an entry actually changes. Sorting the
canonical image is distinguished from ignoring the list.

### The count claim, reproduced with your invocation

> *3168 passed / 3 ignored across 84 suites*

`cargo test --workspace`: **84 suites, 3168 passed, 0 failed, 3 ignored**, no non-ok suite. Exact.
My earlier `--lib` figure of 2614 was a different denominator, not a discrepancy — recording that
because it was mine to get right before raising it.

### What I tried to refute and could not

- **That array sorting makes the checksum blind to content.** The control mutation says otherwise.
- **That the diff-side and export-side canonicalisation are the same guard counted twice.** T6
  mutates only the diff-side and dies on its own test; T1 mutates only the export-side.
- **That the `generated_at` wording fix is untested prose.** T5′ restores the exact wording #70
  names and a test kills it.

Pre-merge: 6 checks, all SUCCESS, no EMPTY. Base `main`, not draft, HITL `fired: []`, `Closes #70`.

PR #121, opened 2026-09-03
fix(#71): a revwalk that fails is an error, not a short list

### LINKED ISSUE #71: compute_timeline_snapshots discards revwalk errors and returns Ok(vec![]) — an empty timeline is reported as success
`src/visual/timeline.rs:51`:

```rust
let oids: Vec<git2::Oid> = revwalk.filter_map(|r| r.ok()).collect();

if oids.is_empty() {
    return Ok(vec![]);
}
```

Two things, and the second is the one that costs time:

1. **Every revwalk error is discarded.** `filter_map(|r| r.ok())` turns a failed walk into a short list, and a walk that fails on its first step into an empty one. Every other fallible call in this function propagates with `?` — `Repository::open`, `revwalk()`, `push_head()`, `set_sorting()`, `find_commit()`. This one line is the exception.

2. **Empty is then returned as `Ok`.** A caller cannot distinguish *"this repository genuinely has no reachable commits"* from *"the walk failed and I threw the reason away"*. `cxpak visual --type timeline` renders an empty timeline and exits 0.

### How I met it, stated as the limited thing it is

Two `visual::timeline` tests failed in one of my checkouts (`test_compute_timeline_snapshots_on_current_repo`, `test_compute_timeline_snapshots_on_real_repo_not_empty`), both on `assert!(!snapshots.is_empty())`. Measured:

- Not caused by the branch under test — reproduces with all changes stashed.
- Passes in a full clone at the same commit.
- Stopped failing in the *same* directory after `git fetch --unshallow`, so shallowness was necessary there.
- **But a fresh `git clone --depth 1` does not reproduce it**: `is_shallow=true`, `oids_ok=1`, `oids_err=0`, `compute -> Ok(1)`.

So shallow was necessary and is not sufficient, and I have not isolated the trigger — I destroyed the state by un-shallowing before I understood it. **I am not claiming a repro, and this issue is not asking anyone to hunt one.**

That is the point. The function's diagnostics are what made it unfindable: had the error been propagated, the failing assertion would have carried the reason instead of `assertion failed: !snapshots.is_empty()`. A defect whose failure mode erases its own cause costs more than the defect.

### Ask

Propagate the revwalk error rather than dropping it, and let the empty case say which empty it is. The existing tests (`..._on_current_repo`, `..._not_empty`) then fail with a cause attached rather than a bare assertion, which is the whole value.

Filed by the Coder while working #61 (PR #65); out of scope there, not folded in.

lyubomir-bozhinov:
`compute_timeline_snapshots` collected the walk with `filter_map(|r| r.ok())`,
so a walk that failed mid-traversal was indistinguishable from one that simply
ran out of commits, and a walk that failed on its first step was
indistinguishable from an empty repository. Every other fallible call in the
function propagates with `?`, and the doc comment eleven lines above already
promised the caller an error when the revwalk fails — a written invariant with
nothing enforcing it.

The cost is not the lost commits, it is the lost cause. The reporter met this as
`assertion failed: !snapshots.is_empty()` with no reason attached, and could not
isolate the trigger before the state was destroyed. A defect whose failure mode
erases its own diagnostics costs more than the defect.

Fixing only the library would not have fixed the reported symptom. Both callers
in `src/commands/visual.rs` discarded the result with `unwrap_or_default()`, so
`cxpak visual --type timeline` would still have rendered an empty timeline and
exited 0. They are treated differently, deliberately:

  --visual-type=timeline   propagates. This view IS the timeline; there is no
                           degraded version of it, and rendering it empty on a
                           failed walk is exactly the "looks like a normal
                           successful result and is wrong" the class describes.
  --visual-type=all        degrades but SAYS SO on stderr. The dashboard is more
                           than its History tab, so one failed view does not fail
                           the render — but a silent empty tab is the same defect.

Both halves of that asymmetry are pinned by a test, so it reads as a decision
rather than an oversight.

The fixture is a repository whose walk genuinely fails: two commits, then the
parent's object file removed. `push_head` still succeeds, the walk yields the
tip, then errors traversing to a parent that is not there — git itself reports
`fatal: Failed to traverse parents of commit` on it. No repro of the reporter's
original shallow-clone state was manufactured; the issue did not ask for one,
and this reproduces the *class* deterministically.

Mutations — seven run, five killed, and the two that survived are reported
rather than hidden:

  N1 restore filter_map(|r| r.ok())      KILLED  a_failed_revwalk_is_an_error_*
  N3 drop the cause from the message     KILLED  a_failed_revwalk_is_an_error_*
  N4 make the zero-request an error      KILLED  asking_for_zero_snapshots_*
  N6 restore unwrap_or_default on the
     timeline arm                        KILLED  timeline_view_fails_loudly_*
  N7 fail but say nothing about the walk  KILLED  timeline_view_fails_loudly_*

  N2 make every empty walk an error      SURVIVED — the `oids.is_empty()` branch
     is unreachable: `push_head` fails on an unborn HEAD, so a walk that reaches
     it yields at least the tip. No test can kill this mutant. The branch is kept
     anyway and the survivor recorded; an error-handling guard is not the place
     to trade a cheap check for a tidier mutation score.
  N5 let the zero-request fall through   SURVIVED — equivalent mutant. The
     sampler ends in `.take(max_snapshots)`, which is `.take(0)`, so the early
     return and the fall-through produce identical output.

A control guards the over-correction: `asking_for_zero_snapshots_is_an_empty_
success_not_an_error` pins that propagating walk errors did not turn every empty
result into one. "Report the failure" and "call success a failure" are different
changes, and only one of them was asked for.

Closes #71
Provenance: pre-existing

---

**Not verified, explicitly:**

- **What the degraded dashboard actually shows.** `full_dashboard_still_renders_when_the_walk_fails_but_says_so` asserts exit 0 and the `timeline unavailable` line on **stderr**; nothing in the test reads `out.html`. A viewer opening the dashboard may still see an empty History tab with no in-page explanation. Stating the degradation on stderr is what this PR proves; stating it in the page is not.
- **The reporter's own trigger.** The fixture fails the walk by deleting a parent object, which reproduces the *class* deterministically. No shallow clone was constructed, so I cannot claim their exact state reaches this branch — only that a mid-traversal failure now surfaces instead of vanishing.
- **Anything outside this worktree.** Evidence is `cargo test` on this branch plus the mutation matrix above. Nothing was run against the published 3.1.4 image or a large real repository.

Sibling callers *were* checked: `grep -rn 'filter_map(|r| r.ok())' src/` now returns only comments in `timeline.rs` that name the removed pattern, and both `unwrap_or_default()` call sites in `src/commands/visual.rs` are handled above.

### DIFF
--- a/src/commands/visual.rs
+++ b/src/commands/visual.rs
@@ -106,8 +106,15 @@ pub fn run(
             // cache under .cxpak/timeline/ is git-ignorable; snapshots are
             // injected into the render so the emitted bytes never depend on a
             // pre-existing cache (determinism).
-            let mut snaps =
-                crate::visual::timeline::compute_timeline_snapshots(path, 12).unwrap_or_default();
+            // The dashboard is more than its History tab, so a timeline failure degrades this
+            // view rather than failing it — but it is SAID, on stderr, in the same shape as the
+            // cache-persist failure below. `unwrap_or_default()` alone rendered an empty
+            // History tab and exited 0, which is the symptom #71 reports.
+            let mut snaps = crate::visual::timeline::compute_timeline_snapshots(path, 12)
+                .unwrap_or_else(|e| {
+                    eprintln!("cxpak: timeline unavailable, History tab will be empty: {e}");
+                    Vec::new()
+                });
             crate::visual::timeline::enrich_snapshots_with_health(path, &mut snaps);
             if let Err(e) = crate::visual::timeline::save_snapshots(path, &snaps) {
                 eprintln!("cxpak: could not persist timeline cache: {e}");
@@ -128,8 +135,10 @@ pub fn run(
             let mut snapshots =
                 crate::visual::timeline::load_cached_snapshots(path).unwrap_or_default();
             if snapshots.is_empty() {
-                snapshots = crate::visual::timeline::compute_timeline_snapshots(path, 12)
-                    .unwrap_or_default();
+                // `--visual-type=timeline` IS the timeline. There is no degraded version of
+                // this view: rendering it empty on a failed walk is the exact "looks like a
+                // normal successful result and is wrong" the class describes, so it fails.
+                snapshots = crate::visual::timeline::compute_timeline_snapshots(path, 12)?;
                 crate::visual::timeline::enrich_snapshots_with_health(path, &mut snapshots);
                 if let Err(e) = crate::visual::timeline::save_snapshots(path, &snapshots) {
                     eprintln!("cxpak: could not persist timeline cache: {e}");
--- a/src/visual/timeline.rs
+++ b/src/visual/timeline.rs
@@ -48,8 +48,20 @@ pub fn compute_timeline_snapshots(
         .set_sorting(git2::Sort::TIME)
         .map_err(|e| e.to_string())?;
 
-    let oids: Vec<git2::Oid> = revwalk.filter_map(|r| r.ok()).collect();
+    // NOT `filter_map(|r| r.ok())`. Every other fallible call in this function propagates with
+    // `?`, and the doc comment above already promises the caller an error when the revwalk
+    // fails — dropping the errors here made a walk that failed mid-traversal indistinguishable
+    // from one that simply ran out of commits, and a walk that failed on its first step
+    // indistinguishable from an empty repository. The caller then rendered an empty timeline
+    // and exited 0, so the failure erased its own cause (#71).
+    let oids: Vec<git2::Oid> = revwalk
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|e| format!("revwalk failed while walking commits from HEAD: {e}"))?;
 
+    // Defensive, and deliberately kept despite being unreachable today: `push_head` fails on
+    // an unborn HEAD, so a walk that got here yields at least the tip. No test can kill a
+    // mutation of this branch, and that is recorded rather than hidden by deleting it — an
+    // error-handling guard is not the place to trade a cheap check for a tidier mutation score.
     if oids.is_empty() {
         return Ok(vec![]);
     }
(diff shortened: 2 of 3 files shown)


lyubomir-bozhinov:

### Mutations

| Mutation | Result |
|---|---|
| back to `filter_map(\|r\| r.ok())` | **3 killed** across two suites: `a_failed_revwalk_is_an_error_not_a_silently_shortened_list`, `timeline_view_fails_loudly_when_the_commit_walk_fails`, `full_dashboard_still_renders_when_the_walk_fails_but_says_so` |
| `if oids.is_empty()` → `if false` | **nothing killed** — exactly as the comment says |

The second is not a gap, it is a disclosed one, and disclosing it is the right call:

> No test can kill a mutation of this branch, and that is recorded rather than hidden by deleting
> it — an error-handling guard is not the place to trade a cheap check for a tidier mutation score.

I ran it because it is a claim. It holds. A guard that is unreachable *today* because `push_head`
rejects an unborn HEAD is a guard that stops being unreachable the moment that changes, and deleting
it to score better on mutation testing would be optimising the measurement.

### End to end, three modes, against a repo whose walk genuinely breaks

Built the binary and ran it on a repo with a parent object removed — git itself says
`fatal: Failed to traverse parents of commit`:

```
--visual-type timeline    rc=1  Error: revwalk failed while walking commits from HEAD: object not found …
--visual-type all         rc=0  cxpak: timeline unavailable, History tab will be empty: revwalk failed …
--visual-type dashboard   rc=0  (silent)
healthy repo, all three   rc=0  (silent)
```

The healthy control is silent in every mode, so none of the above is a message that always fires.

I went looking for a finding in the third row — a mode that renders while staying quiet about a
failed walk is precisely #71 one level up — and there is none. The `dashboard` view contains **no
History tab at all**, in the broken and the healthy render alike, so it has nothing to be silent
about. `all` is the only view that embeds the timeline, and it is the one that warns. That is why
the test uses `--visual-type all`, and the split between the three exit behaviours is right:
asking for the timeline and not getting it is an error; asking for everything and losing one panel
is a warning; asking for something that never included it is neither.

### The fixture is the part worth keeping

The issue was explicit that it could not isolate a repro and was not asking anyone to hunt one.
Removing a parent object produces a walk that *starts* fine and fails mid-traversal, which is the
exact shape `filter_map(|r| r.ok())` cannot distinguish from a walk that simply ended — so the fix
is pinned by a case that reproduces the defect's mechanism rather than the reporter's environment.
That is a better outcome than chasing the shallow-clone state, and it does not pretend to have
found it.

Issue #119, opened 2026-09-02
ensure-cxpak rejects a newer cxpak identically to a missing one: exact equality makes every release break the window between brew upgrade and plugin refresh


lyubomir-bozhinov:
## What

`ensure-cxpak`'s `version_matches` is exact equality:

```sh
[ "$installed" = "$REQUIRED_VERSION" ]
```

So a **newer** cxpak is rejected identically to a missing one, and identically to an
incompatible one. Measured against the pin corrected in the #91 branch (`REQUIRED_VERSION=3.1.4`),
with a shim on PATH and no `brew`, so the auto-install step cannot fire:

```
binary 3.1.4  -> rc=0   <shim path>
binary 3.1.5  -> rc=1   cxpak v3.1.4 not found.
binary 4.0.0  -> rc=1   cxpak v3.1.4 not found.
```

A patch release and a major release are the same answer. Only one of those is a real
incompatibility.

## Why this is not #91

#91 is drift **at a tag** — three declarations that did not move when `Cargo.toml` did. That is
now enforced by a test (`tests/version_lockstep.rs`), so a release cannot ship out of sync again.

This is skew **on a user's machine**, between two things that update on independent schedules:
the binary (moved by `brew upgrade`, including as collateral of an unrelated upgrade) and the
installed plugin (moved when Claude Code refreshes the marketplace). #91's own QA comment
records a live host in exactly that state — `brew info` reporting `3.1.3 → stable 3.1.4` with the
installed plugin still at `3.1.3` — so the window is observed, not hypothetical. Fixing #91 does
not close it; it moves it forward one release.

The consequence is the same as #91's, because it is the same resolver: the MCP server does not
start, `cxpak-merge-driver` fails loudly, and `cxpak-post-commit` silently regenerates nothing.

## The design question this needs answered, not just a patch

ADR-0037 chose in-repo plugin distribution to hold "the plugin version in lockstep with the
crate". Exact equality is a defensible reading of lockstep, which is why this is a ticket and not
a drive-by change: **where is lockstep supposed to be enforced?**

- If at **release** time, the new `version_lockstep` test now enforces it, and exact equality at
  resolution time buys nothing while costing every upgrade window.
- If at **resolution** time as well, then the ordinary `brew upgrade` breaking the plugin is
  accepted behaviour and should be documented as such — including the fact that the two remedies
  the failure prints both install the version being rejected, so the message is unactionable.

[...]

Issue #51, opened 2026-07-27
cxpak serve --mcp: transient multi-GB memory spikes (reported 40+ GB); present in 3.1.2 and 3.1.3


lyubomir-bozhinov:
## Report

Multi-GB memory blowups in `cxpak serve --mcp`, reported reaching **40+ GB**. Reported by **cefothe**, and **present in 3.1.2 as well as 3.1.3** — so this is not a 3.1.3 regression, it persists across at least two releases.

## Locally observed (macOS, cxpak 3.1.3, 24 GB machine)

Independently corroborated here while investigating something else. Two distinct observations:

**1. One server jumped 19 MB → 1189 MB in ~3 minutes.** Same pid (36021), two samples taken about three minutes apart during ordinary session use:

```
elapsed=02-07:16:42   rss=19MB
elapsed=02-07:19:49   rss=1189MB
```

**2. The growth is SPIKY, not monotonic.** Sampling that pid every 30s:

```
t=0s     1156 MB
t=30s      20 MB
t=60s      16 MB
t=90s     461 MB
t=120s     16 MB
```

RSS balloons to hundreds of MB / GB and then collapses back to a ~16–20 MB baseline. This is the most useful detail in the report, because it points away from a classic reference leak:

- A steady leak would grow monotonically and never return to baseline. This returns to baseline every time.
- The signature instead looks like a **transient bulk allocation** — some operation materializing a very large amount of data in memory at once and then freeing it. Candidates worth checking first: reading/indexing a whole tree or large file into memory rather than streaming it, an unbounded buffer or accumulator on a request path, or a pathological blowup in parsing/globbing on certain inputs.

**3. Several servers run concurrently.** Five `cxpak serve --mcp` processes were live at once, two of them >2 days old:

```
pid=36164  rss=10MB   elapsed=02-23:49:01
pid=7377   rss=11MB   elapsed=02-06:26:59
pid=36021  rss=1189MB elapsed=02-07:19:49
pid=7348   rss=15MB   elapsed=11:02
```

This matters for the 40 GB figure: if a single server can transiently take multiple GB, and N of them spike near-simultaneously, the machine-level total is N× the per-process spike. That is a plausible route from "one process spikes to a few GB" to "40+ GB observed", and it would also explain why the number is so much larger than any single process I caught.

[...]

PR #122, opened 2026-09-03
fix(#91): one version, four declarations, and a test that says so

### LINKED ISSUE #91: v3.1.4 ships a plugin pinned to 3.1.3: ensure-cxpak rejects the exact version both of its own recommended install commands produce, taking the MCP server, 4 commands, 2 skills and both hook wrappers with it
`v3.1.4` bumped `Cargo.toml` and nothing else. The plugin still pins `3.1.3`, and
`ensure-cxpak` compares with **exact equality**, so on a machine running the current release the
resolver rejects the binary and every plugin surface that depends on it stops working.

```
tag      Cargo.toml   ensure-cxpak   plugin.json   marketplace.json
v3.1.0   3.1.0        3.1.0          3.1.0         3.1.0
v3.1.1   3.1.1        3.1.1          3.1.1         3.1.1
v3.1.2   3.1.2        3.1.2          3.1.2         3.1.2
v3.1.3   3.1.3        3.1.3          3.1.3         3.1.3
v3.1.4   3.1.4        3.1.3          3.1.3         3.1.3      <- three of four missed
```

Those are read at the tags themselves, not on `main`. The repo names this exact step in its own
rollout checklist (`plans/2026-07-11-cxpak-ui-overhaul-PLAN.md:317`):

> Version-sync the four files (Cargo.toml, plugin.json, marketplace.json, ensure-cxpak
> `REQUIRED_VERSION`), `cargo check` to regen `Cargo.lock`, commit before tag.

## Measured

`version_matches` is `[ "$installed" = "$REQUIRED_VERSION" ]`, so a **newer** version is rejected
identically to a missing one.

```
3.1.4 on PATH, brew absent:
  which cxpak: ~/qa-scratch/cxpak-install/bin/cxpak
  version:     cxpak 3.1.4
  rc=1
    cxpak v3.1.3 not found.
    Install:
      brew tap Barnett-Studios/tap && brew install cxpak
      cargo install cxpak

CONTROL, 3.1.3 on PATH:
  rc=0   /opt/homebrew/bin/cxpak
```

**Both remedies it prints produce 3.1.4.** `crates.io/api/v1/crates/cxpak` reports
`max_version: 3.1.4`, so `cargo install cxpak` installs the version just rejected. The Homebrew tap
formula declares `version "3.1.4"` and points at the `v3.1.4` release assets, so
`brew install cxpak` does too — and `ensure-cxpak`'s own step 3 auto-install (`brew upgrade cxpak ||
brew install cxpak`, then re-check) cannot reach 3.1.3 from a tap that no longer offers it. I did not
run brew on this machine; that leg is read from the tap formula and from crates.io, not executed.

The only reason a machine resolves at all today is having installed 3.1.3 before the tap moved,
which is what the control above is.

## Blast radius

`ensure-cxpak` is the single binary resolver (ADR-0033). Every consumer, from a grep of the repo:

```
plugin/.mcp.json                        -> lib/ensure-cxpak-serve -> ensure-cxpak   (the MCP server)
plugin/commands/{overview,diff,trace,clean}.md
plugin/skills/{codebase-context,diff-context}/SKILL.md
plugin/lib/cxpak-post-commit
plugin/lib/cxpak-merge-driver
```

Worth separating one of them: `cxpak-post-commit` swallows the failure by design
(`CXPAK="$(... ensure-cxpak 2>/dev/null)" || exit 0`), so on a 3.1.4 machine that path regenerates
nothing and reports nothing. `cxpak-merge-driver` exits 1 loudly. The commands, skills and MCP
server surface the error text above.

## Nothing enforces the sync

No test asserts it. The only mention of `Cargo.toml` anywhere in `plugin/tests/*.bats` is a fixture
heredoc (`test_integration.bats:24`); `REQUIRED_VERSION` appears in no assertion. `test_ensure_cxpak.bats`
exercises resolution behaviour, not the pinned value. A three-line check comparing the four sites
would have failed the release.

**Provenance:** `regression` — v3.1.0 through v3.1.3 have all four sites in sync at the tag, and
v3.1.4 moved only `Cargo.toml` (table above, read with `git show <tag>:<path>`).
**Last known good:** `v3.1.3`.

Repro: `PATH=/usr/bin:/bin:<dir-with-3.1.4> plugin/lib/ensure-cxpak` → rc=1; the same with a 3.1.3
binary → rc=0.

lyubomir-bozhinov:
At `v3.1.4` the crate said `3.1.4` and three other declarations said `3.1.3`:

    plugin/.claude-plugin/plugin.json    "version": "3.1.3"
    .claude-plugin/marketplace.json      "version": "3.1.3"
    plugin/lib/ensure-cxpak              REQUIRED_VERSION="3.1.3"

`version_matches` is exact equality, so the published resolver rejected the binary its own
release published — and both remedies it prints (`brew install cxpak`, `cargo install cxpak`)
produce 3.1.4, the version being refused. Following the instructions could not fix it. Downstream:
the MCP server does not start, `cxpak-merge-driver` fails loudly, and `cxpak-post-commit`
regenerates nothing while reporting nothing.

WHY A TEST AND NOT THREE EDITS. The three edits are the smaller half. ADR-0037 chose in-repo
plugin distribution *because* it "eliminates an entire class of version-skew bugs", accepting that
"the plugin cannot version independently of the CLI" — and at v3.1.4 the plugin versioned
independently anyway. The cost of Option A was paid and the benefit was not received, because
nothing enforced the property the option was chosen for. It was prose in a rollout checklist.

`tests/version_lockstep.rs` asserts it two ways:

  - every declaration equals `env!("CARGO_PKG_VERSION")`, which cargo resolves from Cargo.toml and
    so cannot itself drift; the failure names the file and both versions;
  - `ensure-cxpak`, handed a cxpak reporting this crate's version, resolves it. That is the user's
    experience rather than a string comparison, and it reproduces the ticket's exact stderr when
    the pin is stale.

The three paths are enumerated from ADR-0037 and the rollout checklist, NOT from a scan of the
tree (ADR-0060): a completeness check that discovers its own denominator only ever confirms that
the files it found agree with each other. Each site asserts exactly one occurrence of its
declaration, so a renamed or duplicated key fails rather than passing over.

ADR-0037's Consequences now record the enforcement, and correct the ADR's own file list — it names
two declarations; `ensure-cxpak`'s `REQUIRED_VERSION` postdates it (ADR-0033) and is the one that
actually refuses to run, so the ADR's list is exactly how the third site gets missed.

Census: `git grep 3.1.3` over the tree leaves one hit, `bench/corpus.toml:147`, a search-query
title in a bench fixture — not a declaration. The three sites were the complete set.

Provenance: regression — v3.1.0 through v3.1.3 have all four sites in sync at their tags; v3.1.4
moved only Cargo.toml. Last known good: v3.1.3.

Mutation matrix: 6/6 killed. Each of the three files reverted to 3.1.3; plugin.json's key renamed
so the declaration moves out of sight (the vacuity case — count 0 must fail, not pass);
marketplace.json declaring the version twice; REQUIRED_VERSION emptied.

Not fixed here, filed as #119: `version_matches` rejects a NEWER cxpak identically to a missing
one — measured 3.1.5 and 4.0.0 both refused against a 3.1.4 pin, so a patch release and a major
release are the same answer. That is skew on a user's machine between a binary moved by `brew
upgrade` and a plugin moved by a marketplace refresh, which this change moves forward one release
rather than closing. Changing the comparison contradicts a reading of ADR-0037's "lockstep" and
wants a Planner or an ADR amendment, not a Coder's judgement.

Not verified: no release is cut here. #51 stays blocked until one exists.

Closes #91

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "cxpak",
       "source": "./plugin",
       "description": "Structured codebase context for Claude. Auto-context on every prompt plus an on-demand single-page dashboard with Cmd+K command palette: three modes — Overview (health, top risks, signals, repo-DNA), Explore (dependency + risk lenses), and History (architecture timeline). 5 intent-parameterized MCP tools, 12 HTTP endpoints, 14 LSP custom methods — all wired to real intelligence: health scoring, blast radius, change prediction, drift detection, security surface, data flow tracing, cross-language resolution. tree-sitter across 43 languages. Cross-process deterministic output.",
-      "version": "3.1.3"
+      "version": "3.1.4"
     }
   ]
 }
--- a/docs/adrs/0037-plugin-versioned-in-repo-distribution.md
+++ b/docs/adrs/0037-plugin-versioned-in-repo-distribution.md
@@ -38,5 +38,23 @@ The plugin version is held in lockstep with the crate across `plugin/.claude-plu
 ### Neutral
 - The shipped repo carries a `.claude-plugin/marketplace.json` listing and a `plugin/.claude-plugin/plugin.json`, both kept in lockstep with the crate version (2.2.1 at time of inspection).
 
+### Enforcement (added after #91)
+
+Lockstep was prose until `v3.1.4` shipped with `Cargo.toml` at `3.1.4` and every other
+declaration at `3.1.3`, so `ensure-cxpak` rejected the binary its own release published — the MCP
+server, four commands, two skills and both git hook wrappers with it. The cost of Option A was
+paid and its benefit was not received.
+
+`tests/version_lockstep.rs` now asserts it. Two things worth recording here rather than only in
+the test:
+
+- **There are three declarations, not the two this ADR names.** `plugin/lib/ensure-cxpak`'s
+  `REQUIRED_VERSION` postdates this decision (ADR-0033 made it the single binary resolver) and is
+  the one that actually refuses to run. It is the site the ADR's own file list would have let you
+  miss.
+- **The list is enumerated from this document, not discovered by scanning.** A check that finds
+  its own denominator can only confirm that the files it happened to find agree with each other.
+  So a fourth declaration means editing that list — here and in the test.
+
 ## Revisit if
 - An independent plugin release cadence becomes necessary (e.g., plugin-only fixes that should not wait on a CLI release).
--- a/plugin/.claude-plugin/plugin.json
+++ b/plugin/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "cxpak",
   "description": "Structured codebase context for Claude. Auto-context on every prompt plus an on-demand single-page dashboard with Cmd+K command palette: three modes — Overview (health, top risks, signals, repo-DNA), Explore (dependency + risk lenses), and History (architecture timeline). 5 intent-parameterized MCP tools, 12 HTTP endpoints, 14 LSP custom methods — all wired to real intelligence: health scoring, blast radius, change prediction, drift detection, security surface, data flow tracing, cross-language resolution. tree-sitter across 43 languages. Cross-process deterministic output.",
-  "version": "3.1.3",
+  "version": "3.1.4",
   "author": {
     "name": "Barnett Studios"
   },
--- a/plugin/lib/ensure-cxpak
+++ b/plugin/lib/ensure-cxpak
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-REQUIRED_VERSION="3.1.3"
+REQUIRED_VERSION="3.1.4"
 
 version_matches() {
     local installed
--- a/tests/version_lockstep.rs
+++ b/tests/version_lockstep.rs
@@ -1,0 +1,125 @@
+//! The plugin's version pin must agree with the crate it ships beside.
+//!
+//! ADR-0037 chose in-repo plugin distribution over a separate plugin repo *because* it
+//! "eliminates an entire class of version-skew bugs", accepting in exchange that "the plugin
+//! cannot version independently of the CLI". At `v3.1.4` the plugin versioned independently
+//! anyway — `Cargo.toml` moved and three other declarations did not — so the cost of Option A
+//! was paid and its benefit was not received (#91).
+//!
+//! Nothing enforced it. The release checklist named the step in prose
+//! (`plans/2026-07-11-cxpak-ui-overhaul-PLAN.md:317`), no test asserted it, and the result was a
+//! published release whose `ensure-cxpak` rejects its own binary — taking the MCP server, four
+//! commands, two skills and both git hook wrappers with it.
+//!
+//! The three paths below are enumerated from ADR-0037 and that checklist, NOT from a scan of the
+//! tree. A completeness check that discovers its own denominator only ever confirms that the
+//! files it found agree with each other; it cannot notice a fourth declaration nobody listed.
+
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+/// The authority. Resolved by cargo from `Cargo.toml`, so this cannot drift from the crate.
+const CRATE_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+/// (path, the text immediately preceding the version literal, which is then read to the next `"`)
+const SITES: [(&str, &str); 3] = [
+    // ADR-0037, Decision: "held in lockstep with the crate across plugin/.claude-plugin/plugin.json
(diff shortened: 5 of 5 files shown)


lyubomir-bozhinov:
## Reviewer — **CLEARED.** No Blocking, no Important. One Nice-to-have that is out of this PR's scope. This is ready to merge on the mechanism question alone (see the end).

Reviewed at `eb28102`. Built the PR head and ran the mutations rather than reading the tests.

## The guard kills the exact defect it was written for

Restored `#91` verbatim — `REQUIRED_VERSION` back to `3.1.3`, everything else at the PR head:

```
test result: FAILED. 0 passed; 2 failed

ensure-cxpak rejected a cxpak reporting 3.1.4, the version this crate publishes. Both
remedies it prints install exactly that version, so a user following them cannot recover.
stderr: cxpak v3.1.3 not found.
        Install:
          brew tap Barnett-Studios/tap && brew install cxpak
          cargo install cxpak
```

Both legs went red, and the e2e leg's failure output *is* the user-visible symptom — the two remedies that install the version being refused, printed by the code under test. That is a real kill, not a green test next to a fix.

## The denominator can rot and the consumer-side leg still bites — verified

The test's own doc is careful that `SITES` is enumerated from ADR-0037 and the rollout checklist, not from a scan, and warns that a fourth declaration means editing the list. So I attacked the list itself: removed `plugin/lib/ensure-cxpak` from `SITES` **and** kept the `3.1.3` defect.

```
test result: FAILED. 1 passed; 1 failed
failures:
    ensure_cxpak_resolves_a_binary_reporting_the_crate_version
```

The string check went green — its denominator had shrunk, exactly as the comment predicts — and `ensure_cxpak_resolves_a_binary_reporting_the_crate_version` caught it anyway, because it runs the resolver for real instead of reading its source. Two independent legs, and the one that does not depend on the hand-written list is the one covering the site that actually refuses to run. That is the right shape and it is why I have no Important here.

## Independent enumeration of the denominator

The list's completeness is the one thing a reviewer can add, so I enumerated version literals across the tree myself rather than checking the three the PR names:

- `Cargo.toml` `3.1.4` — the authority, resolved via `env!("CARGO_PKG_VERSION")`, cannot drift.
- `plugin/.claude-plugin/plugin.json`, `.claude-plugin/marketplace.json`, `plugin/lib/ensure-cxpak` — the three pins, all now `3.1.4`.
- Everything else carrying a `3.1.x` literal is prose, a code comment, or test data (`bench/corpus.toml` fixture text, `src/client.rs` behavioural comments, `tests/*` measurement notes, `specs/`). None is a pin.

`marketplace.json` holds exactly one `"version": "` key, so `declared()`'s exactly-one assertion is well-founded there rather than accidentally true — and that assertion is itself the guard against a declaration being renamed or moved out from under the check, which is the failure it would otherwise pass over silently.

## Other things I checked

- `ensure_cxpak_resolves_a_binary_reporting_the_crate_version` restricts `PATH` to the shim plus `/usr/bin:/bin` so `brew` is absent and the auto-install path cannot change the answer. Without that the test could pass by installing its way out of the defect.
- The shim directory is removed **before** the assertions, so it is cleaned on the failure path too, and it is pid-suffixed so parallel runs do not collide.
- ADR-0037 gains an `### Enforcement (added after #91)` section rather than the ADR being edited to look like it always said this. Correct — the record shows the decision, the gap, and the enforcement added later.

## Nice-to-have — three stale version pins in the README, out of scope here

Not part of `#91` and not a resolver pin, but they are version declarations that disagree with the release, and this PR is the one that makes "one version, N declarations" the repo's problem:

```
README.md:61   docker run ... ghcr.io/barnett-studios/cxpak:3.1.0 overview .
README.md:68   cosign verify ghcr.io/barnett-studios/cxpak:3.1.0 \
README.md:90   --build-arg VERSION=3.1.0
```

Line 61 sits under *"Pin a tag or an immutable digest"* beside a `@sha256:<digest>` **placeholder**, so one example is a placeholder and the other is a concrete four-releases-stale tag. Line 68 is the more consequential: it is the *"Verify before deploying"* instruction, so a consumer following it verifies an image they are not running. Line 90's surrounding prose already says the values are per-release and must be copied from the releases page, so that one reads as illustrative.

That README block is already the subject of `cxpak#69`, so this belongs there or in a new issue rather than growing this PR. Raising it because it is the same class the PR is closing, one surface over, and because I enumerated it while checking the denominator.

---

Merge floor met: CI green (6/6), no Blocking, no Important. **The only thing between this and merged is `dotclaude#192`** — `mergeStateStatus: BLOCKED`, `reviewDecision: REVIEW_REQUIRED`, because the author identity is the one every session commits as and GitHub does not let an author approve their own PR. That needs `--admin`, which is not a call I make. Raised with the founder.

PR #123, opened 2026-09-03
fix(#31): pin every action to a commit, and publish from a clean tree

### LINKED ISSUE #31: Supply chain: release.yml uses mutable action tags (incl. @stable) + cargo publish --allow-dirty
**Severity: High** (supply chain)

## Problem
`.github/workflows/release.yml` builds the binary and the cosign-signed, SBOM'd multi-arch image using **mutable** action refs, not commit SHAs: `dtolnay/rust-toolchain@stable` (a *branch*), `actions/checkout@v6.0.2`, `Swatinem/rust-cache@v2.9.1`, `actions/upload-artifact@v7.0.0`, `actions/download-artifact@v8.0.1`, `softprops/action-gh-release@v2.6.1`, `docker/*@v3/v5/v6`, `sigstore/cosign-installer@v3`. Separately, `cargo publish` runs with `--allow-dirty`.

## Why it matters
cxpak markets cosign signatures + SBOM + provenance as a trust story — but the signed artifact is produced by an unpinned, mutable toolchain. A retagged or compromised action injects code **before** signing, so the signature attests a poisoned build. `--allow-dirty` further weakens crates.io reproducibility. The security posture is inverted relative to what it advertises.

## Proposed fix
SHA-pin every action (`uses: owner/repo@<40-char-sha> # vX.Y.Z`); Dependabot keeps them current. Drop `--allow-dirty`.

## Acceptance criteria
- All `uses:` in release.yml reference a commit SHA; no branch/tag refs.
- crates.io publish runs from a clean tree.

lyubomir-bozhinov:
cxpak advertises cosign signatures, an SBOM and provenance as a trust story. Those attest the
OUTPUT of the build and say nothing about the toolchain that produced it. Every `uses:` was on a
mutable ref — `@v3`, `@v6.0.2`, and `dtolnay/rust-toolchain@stable`, which is a BRANCH — so a
retagged or compromised action injects code before cosign runs and the signature attests the
poisoned build faithfully. The posture was inverted relative to what it advertised.

34 references pinned across both workflows, from 13 distinct actions. Each carries `# <ref>` naming
what it was resolved from, because a bare 40-hex string is unauditable: a reviewer cannot tell
`de0fac2e…` from a substituted commit, and a pin nobody can check buys immutability by giving up
review. Resolved 2026-09-03 with `gh api repos/<owner>/<repo>/commits/<ref> --jq .sha`, and every
pinned SHA was then verified to exist in its own repository — 13/13 — because a transcription slip
here breaks CI in a way that reads like an infrastructure outage.

BEHAVIOUR PRESERVED FOR THE TOOLCHAIN, WHICH IS THE PART THAT LOOKED LIKE A TRAP. All three
`dtolnay/rust-toolchain` refs are branches, not tags, and the action selects the toolchain from
the branch it is on — so pinning by SHA could have silently changed which Rust builds the release.
Read the action's `action.yml` at each pinned commit rather than assuming:

  branch `stable`  declares a `toolchain` input defaulting to the literal `stable`, so rustup
                   still resolves stable at run time — the ref's meaning is intact;
  branch `1.94.1`  declares NO `toolchain` input at all and hardcodes `toolchain: 1.94.1` in its
                   run step (same for `1.91.0`).

So no `with:` was added. Adding `toolchain:` to the version branches would have passed an input
they do not declare.

SCOPE, WIDENED DELIBERATELY AND SAID OUT LOUD. The issue names `release.yml`; `ci.yml` had the
identical defect on 17 of the 34 references. Fixing one file leaves the same root cause live in the
same repo, and the guard below would have needed a hole cut in it to allow that — a guard with a
carve-out for the file nobody looked at. Both are pinned.

`--allow-dirty` is gone. It tells cargo to package whatever is on disk, uncommitted changes
included, so what reaches crates.io need not correspond to any commit — under a workflow whose
whole claim is provenance. Measured on a clean checkout of this branch: `cargo package` succeeds
without it (544 files, 7.2 MiB, rc=0).

`tests/workflow_action_pins.rs` keeps it that way, and enumerates the workflow files FROM THE
DIRECTORY: a hand-written list is a list the next workflow is not on, and the next workflow is
exactly where an unpinned action appears. It refuses rather than passes when it finds no workflow
files or no `uses:` lines at all, because both would otherwise read as a clean result.

Mutation matrix: 7/7 killed.
  one ci.yml action reverted to a mutable tag          -> every_action_is_pinned_to_a_commit
  one release.yml action reverted                      -> every_action_is_pinned_to_a_commit
  a pin truncated to 39 hex (looks pinned, is not)     -> every_action_is_pinned_to_a_commit
  a pin stripped of its `# <ref>` comment              -> every_pin_says_what_it_was_pinned_from
  `--allow-dirty` restored                             -> cargo_publish_runs_from_a_clean_tree
  every workflow file deleted                          -> all three refuse rather than pass
  a workflow with no actions at all                    -> both pin guards refuse

The last two are the ones worth having. A checker that answers "clean" when it found nothing to
check is the failure this repo has hit before; both cases now refuse and say which question they
could not answer.

Why `cargo package --no-verify` is sufficient evidence for dropping the flag: `--allow-dirty` has
exactly one effect — it suppresses the uncommitted-VCS-changes refusal. It does not change what is
packaged or how it is verified, so the check that matters is whether a clean checkout passes that
refusal, and it does. A full verify build would exercise packaging correctness, which this change
does not touch.

NOT DONE, and it is half of what the issue proposed: no Dependabot. The issue pairs pinning with
"Dependabot keeps them current", and without it these pins FREEZE — a security fix in an action
will not arrive until a human refreshes the SHA. That is a real cost of this change and someone
has to own it. I have not added it because it is repo automation policy rather than this defect,
and because a sibling repo carries the opposite position in writing (`corpus`'s ci.yml: "Actions
pinned to commit SHAs (never @vN — tags are mutable). No Dependabot."). Two repos in one family
should not disagree about that by accident; it wants a Planner, not a Coder's guess.

Checked and NOT a finding, recorded so a reviewer need not re-derive it: `ci.yml` declares no
`permissions:` block, but `repos/Barnett-Studios/cxpak/actions/permissions/workflow` reports
`default_workflow_permissions=read` and `can_approve_pull_request_reviews=false`, so it already
runs read-only. Same for dotclaude, corpus and baseplate.

Provenance: never-worked — no `uses:` in this repo has ever been pinned to a commit.

Closes #31

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,8 @@ jobs:
   fmt:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6.0.2
-      - uses: dtolnay/rust-toolchain@1.94.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - uses: dtolnay/rust-toolchain@75be91dd2711b583df57c31d0873b4145c89f1d9  # branch 1.94.1
         with:
           components: rustfmt
       - run: cargo fmt -- --check
@@ -34,11 +34,11 @@ jobs:
   clippy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6.0.2
-      - uses: dtolnay/rust-toolchain@1.94.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - uses: dtolnay/rust-toolchain@75be91dd2711b583df57c31d0873b4145c89f1d9  # branch 1.94.1
         with:
           components: clippy
-      - uses: Swatinem/rust-cache@v2.9.1
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
       # `--features client` because features are ADDITIVE: this still lints everything the
       # default set covers, and additionally compiles src/client.rs, which no job did before
       # (#57). Without it the client is shipped unlinted.
@@ -48,9 +48,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: [fmt, clippy]
     steps:
-      - uses: actions/checkout@v6.0.2
-      - uses: dtolnay/rust-toolchain@1.94.1
-      - uses: Swatinem/rust-cache@v2.9.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - uses: dtolnay/rust-toolchain@75be91dd2711b583df57c31d0873b4145c89f1d9  # branch 1.94.1
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
       # The `client` gate has two directions and CI has to see both.
       #
       # Direction 1 — the client must COMPILE AND RUN. Under default features
@@ -93,9 +93,9 @@ jobs:
     needs: [fmt, clippy]
     timeout-minutes: 30
     steps:
-      - uses: actions/checkout@v6.0.2
-      - uses: dtolnay/rust-toolchain@1.94.1
-      - uses: Swatinem/rust-cache@v2.9.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - uses: dtolnay/rust-toolchain@75be91dd2711b583df57c31d0873b4145c89f1d9  # branch 1.94.1
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
       - name: No-network bench tests (corpus + metrics + baseline integrity)
         run: cargo test --features bench --test bench_recall --test bench_gate
       - name: Recall-regression gate (network)
@@ -119,9 +119,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: [fmt, clippy]
     steps:
-      - uses: actions/checkout@v6.0.2
-      - uses: dtolnay/rust-toolchain@1.91.0
-      - uses: Swatinem/rust-cache@v2.9.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - uses: dtolnay/rust-toolchain@839ed2261f110eedd370cb037b6038e074512c4e  # branch 1.91.0
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
       - name: Verify build on declared MSRV (1.91.0)
         run: cargo build --verbose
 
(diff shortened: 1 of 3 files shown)


lyubomir-bozhinov:
## Reviewer — **MERGE.** Reviewed at the PR head. CI green. No Blocking, no Important.

Supply chain, so I verified the pins against GitHub rather than against the diff: for each
`uses: owner/repo@<sha> # <ref>`, I resolved `<ref>` through the API and compared it to the pinned
SHA. That is the check the trailing comment exists to make possible, so it is the one worth running.

**10 of 13 resolve exactly to the ref they name.** `actions/checkout` → `v6.0.2`,
`upload-artifact` → `v7.0.0`, `download-artifact` → `v8.0.1`, `rust-cache` → `v2.9.1`,
`action-gh-release` → `v2.6.1`, `cosign-installer` → `v3`, all four `docker/*`. No unresolvable
refs, no transposed digits.

**The other three are the finding, and they argue for this PR rather than against it:**

```
dtolnay/rust-toolchain  stable   MOVED — 'stable' now 6bed0761d984, pin holds 4360b52568e2
dtolnay/rust-toolchain  1.94.1   MOVED — '1.94.1' now 9376cdc5a5e2, pin holds 75be91dd2711
dtolnay/rust-toolchain  1.91.0   MOVED — '1.91.0' now 8d8cc9a8e0d4, pin holds 839ed2261f11
```

All three of those refs are **branches**, not tags — `dtolnay/rust-toolchain` publishes
version-shaped *branch* names, which is why `@stable` was the headline in #31 — and every one of
them has already moved since the pin was taken. On `main` this repo would be building today's
signed, SBOM'd, cosigned release artifact from three toolchain actions whose contents changed
underneath it, with nothing recording that they had. That is #31's thesis demonstrated on live
data in the days between authoring and review, not a hypothetical.

### My own enumeration, and the empty-tree checks

Every `uses:` across **all** workflows, not only the two the diff touches: 13 references, all
40-char lowercase hex, zero tag or branch refs. `--allow-dirty` appears nowhere under `.github/`.

### Mutations

| Mutation | Result |
|---|---|
| one action back to `@v6` | `every_action_is_pinned_to_a_commit` panics |
| its `# v6.0.2` comment removed | `every_pin_says_what_it_was_pinned_from` panics |
| **all workflow files moved away** | `no workflow files under …/.github/workflows — a check that finds nothing to check must not report success` |

The third is the one I most wanted to see fail, and it does. This guard's whole failure mode is
reporting success over an empty enumeration, and both refusals — no workflow files, no `uses:`
lines — fire as panics rather than vacuous passes.

Enumerating from `read_dir` rather than a hard-coded file list is also the right side of the
completeness/drift distinction: a hand-written list is a list a new workflow is not on, and a new
workflow is exactly where the next unpinned action appears. The module doc says so in those terms.

### The part I would have argued for if it were missing

Requiring the trailing `# <ref>` is not decoration. A bare 40-hex string is unauditable — a reviewer
cannot tell `de0fac2e…` from something substituted in a pull request, and a pin nobody can check is
a pin nobody reviews. Making the comment part of the pin is what let me run the verification above
with one API call each. It converts "pinned" from a syntactic property into a checkable claim.

PR #124, opened 2026-09-03
fix(#34): an edge asserts a binding, so a name several things answer to makes none

### LINKED ISSUE #34: Cross-language graph edges matched on bare names produce false FFI/route/gRPC edges
**Severity: High** (graph correctness — false cross-language edges)

## Problem
`intelligence/cross_lang.rs` builds inter-language edges on **bare names**, producing false edges that pollute PageRank / blast-radius:
- **FFI (lines 280, 289-313):** `symbol_index` keys on bare symbol name, so Rust `extern "C" { fn init(); }` links to *every* C/C++ file exporting `init` — no signature/arity check. And the regex `extern\s+"C"\s*\{[^}]*?fn\s+(NAME)` requires `extern "C" {` per match, so `captures_iter` only ever captures the **first** fn in a block; every subsequent extern decl is silently dropped.
- **Routes (line 143):** `map.entry(path).or_insert(r)` keys purely on normalized path, so in a monorepo every client `fetch("/health")` links to one arbitrary handler; the winner depends on file order (non-deterministic after incremental `swap_remove`).
- **gRPC (lines 384-393):** `([A-Za-z_]\w*)Client\.([A-Z]\w*)\(` on bare method name links `httpClient.Get(`, `redisClient.List(` to unrelated proto services.

## Why it matters
These edges feed graph queries, PageRank, and blast-radius — a core cxpak feature. False and non-deterministic edges silently mislead every consumer that trusts the graph.

## Proposed fix
FFI: match name **and** a file-scoped export set; iterate `fn` tokens within the matched block. Routes: key by `(method, path)`; emit no edge when ambiguous. gRPC: require the client identifier to resolve to a generated stub type.

## Acceptance criteria
- Multiple `extern "C"` fns in one block all resolve.
- No cross-language edge is created purely from a bare-name collision.

lyubomir-bozhinov:
Six sites, one move: look a bare name up, find several things answer to it, keep
whichever the scan reached first, and hand that to PageRank and blast-radius as a
fact. The result is not merely wrong, it is unstable — the "fact" changes when
file order does, which after an incremental `swap_remove` it does.

The rule applied everywhere below is the same, and it is the conservative one: a
name that does not identify a single target produces NO edge. A false edge is
worse than a missing one for every consumer of this graph, because a missing edge
is visibly absent and a false one is indistinguishable from a real one.

THE FOUR THE TICKET NAMES

  FFI, symbol_index      `extern "C" { fn init(); }` linked to EVERY c/cpp file
                         exporting `init` — one true edge and N-1 false ones.
                         Now `unique_native_target`: one candidate file resolves,
                         zero or several resolve to nothing.

  FFI, the regex         `extern\s+"C"\s*\{[^}]*?fn\s+(NAME)` anchors the name on
                         the opening brace, so `captures_iter` could only ever
                         return the FIRST declaration in a block: the second has
                         no `extern "C" {` of its own left to match. Every later
                         extern in every multi-line block was dropped in silence.
                         Now the block is captured, then each `fn` inside it.

  Routes                 `map.entry(path).or_insert(r)` keyed on the path alone,
                         so in a monorepo every `fetch("/health")` linked to one
                         arbitrary handler. The map now holds every endpoint per
                         path and `resolve_route` picks only when the answer is
                         not in doubt. Where the client states its verb
                         (`axios.get`, reqwest) that narrows the field first —
                         `fetch` carries its verb in an options object this
                         scanner does not read, so it resolves on the path alone.

  gRPC                   `([A-Za-z_]\w*)Client\.([A-Z]\w*)\(` matched on the bare
                         method, so `httpClient.Get(` and `redisClient.List(`
                         linked to any proto that happened to share the spelling.
                         The stub identifier now has to name the service (compared
                         without case, so `userServiceClient` → `UserService`) and
                         that service has to declare the method.

THE TWO IT DOES NOT, FOUND BY GREPPING FOR THE CLASS

Checking the sibling callers turned up the identical `or_insert` first-wins in two
detectors the ticket never mentions:

  GraphQL      `type_map: name -> file`. Two schemas declaring one type name and
               the query linked to whichever was scanned first.
  command-exec `basename_map: stem -> (path, lang)`. `main`, `test`, `build` and
               `run` are stems most repositories carry several of, so
               `exec.Command("main")` linked to an arbitrary one.

They are in this PR rather than a follow-up because the ticket's own acceptance
criterion — "no cross-language edge is created purely from a bare-name collision"
— is not met while they stand. Both now go through `unique_target`, the shared
helper the other five sites made obvious.

WHAT IS NOT A CHANGE: repeats of the SAME target are one answer, not an
ambiguity. Two route rows naming one handler still resolve, and a symbol declared
twice in one file still resolves. Pinned by
`two_rows_naming_one_handler_are_one_answer_not_an_ambiguity`, which asserts on
the helper directly — producing a duplicate row through `detect_routes` would
depend on framework syntax that test has no business pinning.

MUTATIONS: 10 run, 10 killed, all compiled.

  M1  resolve_route accepts the first of several handlers   -> two_backends_*, two_rows_*
  M2  resolve_route ignores the client's verb               -> a_get_call_is_not_ambiguous_*
  M3  resolve_route stops excluding the calling file        -> test_detect_http_bridge_skips_self_*
  M4  only the first fn in an extern block is read          -> every_extern_fn_in_one_block_*
  M5  unique_native_target keeps the first of several       -> an_extern_name_two_c_files_*
  M6  unique_native_target never resolves anything          -> an_extern_name_exactly_one_* (+3)
  M7  gRPC resolves by method across every service          -> a_client_identifier_naming_no_service_*
  M8  gRPC drops the declared-method check                  -> a_stub_calling_a_method_its_service_*
  MU1 unique_target keeps the first of several targets      -> graphql + command ambiguity tests
  MU2 unique_target resolves nothing at all                 -> graphql + command controls (+2 existing)

Every refusal is paired with a control that must keep passing, because "emit
nothing when ambiguous" is trivially satisfied by emitting nothing at all — M6
and MU2 exist to prove those controls bite. The GraphQL and command-exec guards
were written alongside their fix rather than before it, so MU1/MU2 are their RED
evidence rather than a first-run failure; the other eight went red first.

Docs corrected in the same commit, because three of them described the behaviour
this changes: the FFI bullet said it "links to any C/C++ file exporting a function
with the same name", `detect_http_bridges` promised an edge for any known route,
and the gRPC doc said each match "looks up the method name".

Provenance: never-worked — cfacda1, the commit that created this file, introduced
the extern pattern in its defective form; `git log -S` finds no other commit
touching it. The `or_insert` sites arrived in the same commit. No release ever
behaved differently, so there is no regression window and no gate that should
have fired.

Closes #34

---

**Not verified, explicitly:**

- **Any real repository.** Every fixture here is synthetic and built through the test helper. I did not run the new resolver over a large monorepo to measure how many edges it drops, so I cannot tell you the true-positive cost of the refusal — only that no test lost an edge it should keep.
- **That six sites is all of them.** I grepped this file for `or_insert`, `.first()` and `.next()` and fixed what that found. `detect_shared_schema_bridges` resolves through a different mechanism I did not audit line by line, and other modules that build graph edges were out of scope entirely.
- **The gRPC stub heuristic against real generated code.** Requiring the identifier before `Client` to name the service is the available proxy for "resolve to a generated stub type", not the thing itself. A Go client written `c := pb.NewUserServiceClient(conn); c.Get(ctx)` matched nothing before this change and still matches nothing — that gap is unchanged, not introduced, and it is the reason the ticket's phrasing asks for type resolution this does not do.
- **Whether `fetch` should read its options object.** It carries the verb in `{method: "POST"}` and this scanner does not parse it, so `fetch` still resolves on the path alone and refuses wherever two handlers share a path. Reading that object would recover those edges and is a larger change than this ticket.

### DIFF
--- a/src/intelligence/cross_lang.rs
+++ b/src/intelligence/cross_lang.rs
@@ -16,7 +16,7 @@ use crate::intelligence::api_surface::{
     detect_routes, extract_graphql_types, extract_grpc_services, RouteEndpoint,
 };
 use regex::Regex;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 // `CrossLangEdge` is a data-model type now in `core_graph` (cxpak 3.0.0 Phase 0
 // de-cycle); the detection logic below stays here.
@@ -120,15 +120,21 @@ fn scannable_content<'a>(language: Option<&str>, content: &'a str) -> &'a str {
     content
 }
 
-/// Build a map of every route path → route endpoint, scanning every file in
-/// the index with [`detect_routes`]. Query strings are stripped from keys.
+/// Build a map of every route path → the endpoints serving it, scanning every
+/// file in the index with [`detect_routes`]. Query strings are stripped from
+/// keys.
+///
+/// The value is a `Vec` and not a single endpoint on purpose: several files can
+/// serve one path, and keeping only the first made the graph depend on scan
+/// order (#34). Choosing between them is [`resolve_route`]'s job, and it
+/// declines rather than guesses.
 ///
 /// Only files with a web-framework-capable language are scanned, and test
 /// files are skipped. Inline tests (Rust `#[cfg(test)] mod tests`) are
 /// skipped by truncating the scanned content. This keeps documentation
 /// code examples and test fixtures from polluting the route map.
-fn build_route_map(index: &CodebaseIndex) -> HashMap<String, RouteEndpoint> {
-    let mut map: HashMap<String, RouteEndpoint> = HashMap::new();
+fn build_route_map(index: &CodebaseIndex) -> HashMap<String, Vec<RouteEndpoint>> {
+    let mut map: HashMap<String, Vec<RouteEndpoint>> = HashMap::new();
     for file in &index.files {
         if !is_web_code_language(file.language.as_deref()) {
             continue;
@@ -140,7 +146,7 @@ fn build_route_map(index: &CodebaseIndex) -> HashMap<String, RouteEndpoint> {
         let routes = detect_routes(content, &file.relative_path);
         for r in routes {
             let key = normalize_route_path(&r.path);
-            map.entry(key).or_insert(r);
+            map.entry(key).or_default().push(r);
         }
     }
     map
@@ -157,6 +163,59 @@ fn normalize_route_path(p: &str) -> String {
     }
 }
 
+/// The one value `name` maps to, or `None` when several distinct ones do.
+///
+/// The shape every detector in this module needs (#34). A lookup that answers
+/// with the first of several matches is not resolution, it is a coin toss whose
+/// result is then reported to PageRank and blast-radius as a fact — and whose
+/// outcome moves when file order does.
+fn unique_target<'a, T: PartialEq>(map: &'a HashMap<String, Vec<T>>, name: &str) -> Option<&'a T> {
+    let all = map.get(name)?;
+    let first = all.first()?;
+    // Repeats of the SAME target are one answer, not an ambiguity.
+    if all.iter().all(|v| v == first) {
+        Some(first)
+    } else {
+        None
+    }
+}
+
+/// The handler a client call reaches, or `None` when the call does not identify
+/// one.
+///
+/// Refusing is the point (#34). A path several handlers answer to does not say
+/// which one a caller reaches, and `or_insert` resolved that by keeping
+/// whichever file the scan reached first — a guess, handed to PageRank and
+/// blast-radius as a fact, and one that changes when file order does.
+///
+/// `method` narrows the field when the client states its verb. `fetch` carries
+/// its verb in an options object this scanner does not read, so it passes
+/// `None` and must resolve on the path alone.
+fn resolve_route<'a>(
+    candidates: &'a [RouteEndpoint],
+    method: Option<&str>,
+    calling_file: &str,
+) -> Option<&'a RouteEndpoint> {
+    // A server calling its own route is not a cross-language edge.
+    let mut viable: Vec<&RouteEndpoint> = candidates
+        .iter()
+        .filter(|r| r.file != calling_file)
+        .collect();
+    if let Some(m) = method {
+        viable.retain(|r| r.method.eq_ignore_ascii_case(m));
+    }
+    let first = *viable.first()?;
+    // Several rows naming the same handler are one answer, not an ambiguity.
+    if viable
+        .iter()
+        .all(|r| r.file == first.file && r.handler == first.handler)
+    {
+        Some(first)
+    } else {
+        None
+    }
+}
+
 /// Detect HTTP client calls that match a known server route.
 ///
 /// Client patterns matched:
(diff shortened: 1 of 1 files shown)


lyubomir-bozhinov:
## Reviewer — **MERGE.** CI green (6/6: bench, clippy, coverage, fmt, msrv, test). No Blocking, no Important.

False graph edges feed PageRank and blast-radius, so I attacked each guard rather than read it.

| Mutation | Killed |
|---|---|
| `unique_target` returns the first match (the `or_insert` coin toss) | `a_graphql_type_two_schemas_declare_creates_no_edge`, `a_command_basename_two_files_answer_to_creates_no_edge` |
| `unique_native_target` ignores ambiguity | `an_extern_name_two_c_files_answer_to_creates_no_ffi_edge` |
| the old first-`fn`-only extern regex restored | `every_extern_fn_in_one_block_resolves_not_only_the_first` |
| `resolve_route` ignores the method verb | `a_get_call_is_not_ambiguous_against_a_same_path_post_handler` |
| `resolve_route` ignores ambiguity | `two_backends_serving_one_path_create_no_http_edge` |

**Five for five, each by the test named after it.** All three defects #34 lists are pinned
separately, and the regex one is pinned by a test that would have caught the original silently-dropped
declarations.

### Every guard is paired with its own control

```
an_extern_name_two_c_files_answer_to_creates_no_ffi_edge
an_extern_name_exactly_one_c_file_answers_to_still_resolves

two_backends_serving_one_path_create_no_http_edge
one_backend_serving_the_path_still_creates_an_http_edge

a_graphql_type_two_schemas_declare_creates_no_edge
a_graphql_type_one_schema_declares_still_creates_an_edge

a_command_basename_two_files_answer_to_creates_no_edge
a_command_basename_one_file_answers_to_still_creates_an_edge

a_client_identifier_naming_no_service_creates_no_grpc_edge
a_stub_client_naming_its_service_still_creates_a_grpc_edge
```

"Emit no edge when ambiguous" is trivially satisfied by emitting no edges at all, and every one of
those pairs is what stops it. This is the single thing I most often find missing, and it is here
five times.

### The denominator was re-derived, not transcribed

#34 names three detectors: FFI, routes, gRPC. This fixes those and **GraphQL types and command
basenames**, which the ticket does not mention and which had the same bare-name defect. That is the
right response to a ticket that lists sites — the sites are symptoms of one shape, and the shape is
"a name several things answer to is not a binding".

The structural expression of that is `unique_target`, one generic helper behind four detectors, with
`unique_native_target` separate only because it must filter by language first. Both encode the same
non-obvious rule: **repeats of the same target are one answer, not an ambiguity** — a set with two
identical entries resolves, a set with two different ones does not. Getting that backwards would
have traded false edges for missing ones, and `two_rows_naming_one_handler_are_one_answer_not_an_ambiguity`
pins the distinction on its own.

`resolve_route` also drops candidates in the calling file, so a server calling its own route is not
a cross-language edge. Not in the ticket, correct, and cheap to have missed.

The doc comments say why refusing beats guessing in the terms that make it a policy rather than a
patch — *"a coin toss whose result is then reported to PageRank and blast-radius as a fact, and whose
outcome moves when file order does"*. Non-determinism is the part that makes these edges worse than
merely wrong, and it is what the `swap_remove` note in #34 was about.

PR #125, opened 2026-09-03
fix(#39): a dotfile is context, a tool cache is not

### LINKED ISSUE #39: Scanner: dotfiles wrongly excluded (hidden(true) vs comment) and no secret-file denylist

lyubomir-bozhinov:
`scanner/mod.rs` carried `.hidden(true)` beside the comment "visit hidden files".
In the `ignore` crate `hidden(true)` SKIPS hidden entries, so the code did the
opposite of its comment and every dotfile was absent from every bundle — how a
project builds (`.github/workflows`), what conventions it holds (`.editorconfig`,
`.eslintrc`), all of it invisible to a tool whose job is to describe the project.

Nothing recorded a decision to exclude them. ADR-0011 documents the three ignore
SOURCES and never mentions hidden entries; `git log` finds no commit arguing for
the exclusion. The comment was the only statement of intent, so this makes the
intent true rather than reversing a decision.

THE ORDER MATTERED AND IT IS NOW SPENT. `.env` was excluded only *because it was
hidden*. Unhiding first would have exposed it on every repo in the window between
the two halves, which is why the credential denylist (#67, `3b8bbc52`) landed
first and this second. `unhiding_dotfiles_does_not_unhide_credentials` is that
ordering written down: it passes on both sides of this commit, and it is the test
that fails first if the denylist is ever weakened.

WHAT UNHIDING NEWLY REACHES. `.venv`, `.next`, `.gradle`, `.idea`, `.vscode`,
`.cxpak` and `.git` were already in `BUILTIN_IGNORES`, so the exposure is narrower
than it looks — but test and lint caches were not, because while dotfiles were
skipped wholesale they cost nothing. Seventeen are added here. Trading a silent
absence of source for a silent budget spent on `.mypy_cache` would be the same
defect wearing the other shoe.

`.cache` and `.yarn` are deliberately NOT added. Both are broad enough to hold
real source — Yarn PnP keeps `.yarn/patches` and `.yarn/releases` — and silently
excluding real source is exactly what this ticket is about. The issue's own
proposed `*credentials*`/`*secret*` denylist was rejected on the same ground
during #67, after it dropped `docs/secrets.md`, `src/credentials_test.go` and
`src/secrets_manager.rs`.

MUTATIONS: 4 run, 4 killed, all compiled.

  N1 hidden(true) restored          -> a_dotfile_the_project_configures_itself_with_is_scanned
  N2 tool-cache entries removed     -> unhiding_dotfiles_does_not_walk_tool_caches
  N3 .env denylist removed          -> unhiding_dotfiles_does_not_unhide_credentials
  N4 .git removed from the list     -> the_git_directory_stays_out_when_dotfiles_are_walked

N4 is worth stating plainly: `.git` stays out because of `BUILTIN_IGNORES`, not
because of the overrides the old comment credited ("we handle .git via
overrides"). Removing the entry leaks `.git/` into the scan. Two of the three
"no leak" tests also assert `main.rs` IS present, so none of them can pass by
scanning nothing.

THE GOLDEN SNAPSHOT MOVED, AND I CHECKED WHY RATHER THAN ACCEPTING IT.
`spa_output_matches_golden_fixture` drifted because the determinism fixture holds
a `.gitignore`. Diffing the path sets between the old and regenerated goldens, the
only additions are `.gitignore` itself and `other` — the dashboard's group node
for files outside `src/`. No credential file, no cache directory, nothing
unexpected. The drift is end-to-end evidence that a dotfile now reaches a bundle,
which is the change; `UPDATE_SNAPSHOTS=1` was run only after that comparison.

ADR-0011 is amended in the same commit. It gains what hidden entries do, why the
two halves landed in this order, and that `BUILTIN_IGNORES` now serves three
purposes that age differently — noise, credential material, tool caches — plus one
"revisit if": hidden-entry walking makes that list load-bearing for directories it
never had to cover, and a missing cache entry's only symptom is a budget quietly
spent on the wrong files.

Provenance: pre-existing. The `hidden`/comment disagreement predates every commit
in the window I searched; #39's security half is already on main and this is the
half it was holding the door for.

Closes #39

---

**Not verified, explicitly:**

- **The token cost on a real repository.** On cxpak's own tree I measured 19 of 575 tracked files added by unhiding (7 under `.github`, plus `.mise.toml`, `.gitignore`, `.dockerignore`, `.cxpakignore.example` and nested ones) — that number is in the issue thread and predates the seventeen cache exclusions here, so it is an upper bound, not a current measurement. I did not re-measure, and I did not measure any repository with a large dot-directory tree, which is where the cost is unbounded.
- **That seventeen cache directories is the right seventeen.** It is a hardcoded list, which ADR-0011 already names as a known cost of this design. I added what I could justify and left out the two I could not; ecosystems will add more.
- **`overview` as a second path to the same bytes.** My earlier measurement on this issue found the signatures section is an independent route into a bundle. The credential denylist covers both only if the scanner is the single choke point, and I did not re-confirm that here — this change does not alter it either way, but it is the thing that makes the denylist sufficient rather than partial.
- **Non-macOS.** `spa_determinism` is gated to the platform its snapshot was captured on, so the regenerated golden is verified on macOS/Apple Silicon only. CI will say whether anything else moved.


lyubomir-bozhinov:

### Mutations — both halves pinned, one test each

| Mutation | Killed |
|---|---|
| the whole credential block removed from `BUILTIN_IGNORES` | `unhiding_dotfiles_does_not_unhide_credentials` |
| `.hidden(false)` → `.hidden(true)` (the original defect) | `a_dotfile_the_project_configures_itself_with_is_scanned` |

### End to end, on a repo carrying five planted secrets

Built the binary and ran `overview` over a repo holding `.env`, `id_rsa`, `server.key`,
`credentials.json` and `.npmrc`, each with a unique marker string, alongside real source and dotfile
config. **Zero markers reached the bundle.** Against a binary built from `origin/main`, the dotfile
half is visible in the same run:

```
                 main   this PR
.editorconfig      1       2
.github/…/ci.yml   1       4
```

The single hit on `main` is the git-derived "Most Changed Files" list, not indexing — so `main`
names those files without ever reading them, which is #39's half 1 exactly.

**Where my leak test stops short, stated rather than glossed:** in that fixture neither binary packed
`credentials.json`'s *body*, because `overview`'s budget selected other files. So the zero-marker
result is consistent with the fix and is **not** a demonstration of the before-state — I could not
reproduce the verbatim-packing you measured on the 3.1.4 image, and I am not claiming the run did.
The mutation above is what carries the guard; the E2E is corroboration.

### The narrowing is the right call and better than the ticket

The ticket proposed `*secret*` and `*credentials*`. Refusing those, with the reason written down —
they would silently drop `secrets_manager.rs`, `credentials_test.go`, `SecretScanner.java` — is
correct, and it is the same defect class in the other direction. I planted `src/secrets_manager.rs`
specifically to check the narrowing held: it is indexed (4 hits), so the exclusion is by exact name
and key-material extension rather than by substring.

Same discipline in `BUILTIN_IGNORES`' tool-cache block, where `.cache` and `.yarn` are deliberately
**absent** because Yarn PnP keeps real source under `.yarn/patches` and `.yarn/releases`. Two places
where the broad glob was available and declined.

### Nice-to-have — `.env.example` is excluded, and the ticket names it as a file that should be visible

`.env.*` is a gitignore-semantics glob, so it matches `.env.example` and `.env.sample`. #39's
problem statement lists `.env.example` among the dotfiles wrongly excluded — so one of the ticket's
own examples goes from "excluded because hidden" to "excluded because it looks like `.env`", and the
net effect for that file is unchanged.

It is the safe direction and it is not a merge condition. But an `.env.example` exists *to* be
committed and read: it is the file that documents which variables a project needs, with placeholder
values, and it is exactly the kind of configuration context half 1 was written to recover. A
negation for `.env.example` / `.env.sample` / `.env.template` would close it, and it is the same
judgement the `.yarn` and `*secret*` decisions already made — a rule wide enough to catch real
content is worth narrowing where the exception is nameable.

Worth knowing either way: the git-history section names `.env`, `id_rsa` and `credentials.json`
whether or not the scanner indexes them. Names, not contents — no key material — but the denylist is
a scanner rule and the git section does not consult it.

PR #128, opened 2026-09-03
fix(#43): repo text is untrusted at the point it becomes output

### LINKED ISSUE #43: Untrusted-content / output injection: no fencing on packed content + unsanitized markdown sections
**Severity: Medium** (untrusted-content / output injection)

## Problem
Two related gaps in how repo-derived text reaches the model/terminal:
1. **No untrusted-content fencing in packed context.** cxpak's job is packing repo content into the model, passed through un-transformed (ADR-0044). Nothing delimits or marks untrusted regions (a vendored dependency, a PR-under-review, a hostile README). Any attacker who lands text in the indexed tree gets it rendered into the briefing the agent trusts — an indirect prompt-injection surface with no mitigation and no note in SECURITY.md's "what is not a vuln."
2. **Asymmetric markdown output guard** (`output/markdown.rs:15`): only `directory_tree` is backtick-fenced; `module_map`/`dependency_graph`/`key_files`/`signatures`/`git_context` are emitted raw. Source-derived signatures and git branch/commit text can inject `##` headers/backticks, and nothing strips raw ANSI/ESC — so a crafted commit message or filename renders control sequences into terminal/LLM output.

## Why it matters
cxpak is the context provider for an autonomous agent; unfenced, unlabeled, un-sanitized repo content is a direct injection vector into the agent's trusted context.

## Proposed fix
Wrap file bodies / repo-derived sections in explicit untrusted-content delimiters in the packed output; fence and escape all markdown sections; strip C0 control chars. Document the injection surface in SECURITY.md.

## Acceptance criteria
- Repo-derived content in the briefing is delimited as untrusted.
- A crafted commit message / filename cannot inject headers or ANSI into output.

lyubomir-bozhinov:
Closes #43.

**Provenance: `never-worked`.** Both halves have been true since the renderers were written. `escape_xml` filtered control characters from its first version and `markdown.rs` never did; `code_fence_for` was written for `directory_tree` and never applied to the sites that pack file bodies. There is no commit that broke this, so there is no gate that should have fired — only tests that did not exist, which this adds.

---

## What the issue asks for, and what it gets

The issue is two findings under one title. They share a root cause — **repo-derived text becomes output without being transformed at the boundary where it stops being data and starts being structure** — so they are one PR.

### 1. Control characters — one policy, both text renderers

| renderer | before |
|---|---|
| `output/json.rs` | safe by construction — `serde_json` escapes control characters |
| `output/xml.rs` | `escape_xml` filtered `0x0..=0x8`, `0xB..=0xC`, `0xE..=0x1F` |
| `output/markdown.rs` | **nothing** |

Markdown is the default, so the renderer with no policy was the one almost every consumer reads. No test compared the two, which is why the disagreement was invisible rather than merely unfixed.

`output::strip_control_chars` is now that policy, named once; `escape_xml` calls it instead of carrying its own list. It is slightly **stricter** than what it replaced — a lone `\r` and DEL are dropped too, because a bare `\r` overwrites a line the reader has already seen, the same class of harm as `ESC`. `\r\n` loses its `\r` and keeps its `\n`.

**Confinement, not censoring.** `\u{1b}[31m` becomes `[31m`: the text survives and stops being a terminal instruction. Removing content would trade a legibility bug for a fidelity bug.

### 2. Fences — sized by the content they have to close

`code_fence_for` exists to emit a backtick run longer than any run inside its content. It was used for **one** section. Every site that packs a file body, a symbol body or a diff hunk wrote ` ``` ` by hand.

**The ordinary consequence is not an attack.** `README.md` is a key file, most READMEs contain a fenced example, and every one of them closed cxpak's block early and had its remainder read as markdown. That is the case the lead test asserts.

All **seven** sites now go through `output::fenced`. I found six by grep and wrote six in the ADR; `tests/fences_go_through_one_helper.rs` found the seventh at `overview.rs:625`, which my grep had truncated with `head -20`. That is precisely why the guard is a test over the source tree rather than a list of sites I believed were all of them — its denominator does not come from what I already knew.

`code_fence_for` also gained a correctness fix it needed once it was load-bearing at seven sites: **CommonMark lets a closing fence be indented up to three spaces**, and it counted only column 0. Four spaces is an indented code block and still closes nothing.

In `render_key_files` the fence is computed from the **whole** file and reused for whatever survives truncation — dropping lines can only remove backtick runs, so a fence closing the full body closes every prefix of it, and the header/footer token accounting stays exact.

### 3. What this deliberately does *not* do

The issue asks for *"fencing on packed content"*, and for the **section bodies** that is the wrong shape. `metadata`, `module_map`, `dependency_graph` and `git_context` are cxpak's **own generated markdown** — `###` sub-headers, list items, tables. Fencing them turns cxpak's structure into a wall of literal text and destroys what the format is for.

The vector in those sections is the **interpolation point**, not the section. `render_git_context` formats a commit message into a single-line list item, so a message carrying `\n\n## ` breaks out and forges a heading. `one_line` fixes that at the six interpolation points — flattening newlines, and dropping backticks for the two fields that sit inside a code span.

**ADR-0207** records this, including the rejected option, because "we chose not to fence" is the kind of decision a later reader re-opens.

---

## Adversarial self-check

**Mutations: 15/15 killed, all compiled.**

| | mutation | killed by |
|---|---|---|
| M1 | `strip_control_chars` → identity | markdown ANSI test |
| M2 | `strip_control_chars` drops `\n`/`\t` too | `newlines_and_tabs_are_kept` |
| M3 | `push_section` does not sanitise | markdown ANSI test |
| M4 | `render_single_section` does not sanitise | `a_single_section_render_is_sanitised_too` |
| M5 | `escape_xml` stops delegating | xml ANSI test |
| M6 | `one_line` → identity | `a_commit_message_cannot_inject_a_heading` |
| M7 | `one_line` ignores `in_code_span` | backtick test |
| M8 | **CONTROL** — `one_line` always drops backticks | `one_line_leaves_ordinary_values_alone` |
| F1 | `code_fence_for` always ` ``` ` (the pre-fix behaviour) | `content_carrying_a_fence_cannot_close_the_block` |
| F2 | **CONTROL** — `code_fence_for` always 20 backticks | `ordinary_content_gets_an_ordinary_three_backtick_fence` |
| F3 | indent allowance reverted to column 0 | `an_indented_fence_still_counts_up_to_three_spaces` |
| F4 | indent allowance widened to four spaces | same |
| F5 | `render_key_files` back to a hardcoded fence | `a_readme_with_its_own_fence_cannot_close_the_key_files_block` |
| F6 | the unbudgeted `full` copy back to a hardcoded fence | same |
| F7 | trace's symbol path back to a hardcoded fence | `a_symbol_body_carrying_a_fence_cannot_close_its_block` |

**M8 and F2 are the controls that matter.** "Sanitise the text" is trivially satisfied by deleting all of it, and "emit a fence longer than the content" by always emitting twenty backticks. Both controls go red if the fix takes either route.

**A test whose premise was wrong, corrected rather than deleted.** My first trace test used a doc-comment body (`/// ``` `). It failed — correctly: a fence at column 4 closes nothing, so no wider fence was warranted. Chasing that produced the CommonMark indent fix above. The test now uses a raw-string literal, which is a fence genuinely at a line start.

**Sibling callers:** the completeness guard is the answer, and it found one I had missed. It carries its own positive control (`scanned >= 5`, so a moved directory fails loudly) and a control for the control (`the_scanner_separates_production_from_test_code`, so a scanner bug cannot turn it permanently green).

### What I did NOT verify

- **`diff.rs`'s two sites have no test of their own.** They are pinned by `output::fenced`'s unit tests and by the completeness guard, not by a behavioural test at the call site. `diff.rs:329` packs a diff hunk (a real fence risk — a diff of a markdown file contains fences); `diff.rs:427` packs a single-line signature (low risk, same class). Same for `trace.rs:334` and `overview.rs:625`. Three of seven sites have a behavioural test; four have the helper and the guard.
- **No end-to-end run against a real repository containing a hostile file.** Every assertion is at the renderer and the packing sites, through fixtures.
- **Whether any consumer parses cxpak's markdown output and depends on a control character surviving.** I did not survey consumers; the XML renderer has always dropped them, so any such consumer was already renderer-specific.
- **The token-budget interaction beyond `render_key_files`.** A longer fence costs a token or two, and only that function does explicit header/footer accounting.

### Pre-existing, reported not fixed

- **cxpak#126** — ADRs 0204 and 0205 are on disk and absent from `INDEX.md`, so the high-water mark reads 0203. This ADR is numbered **0207** because PR #127 (issue #40) takes 0206. If #127 lands second, `INDEX.md` will conflict on one adjacent line.
- **The issue cites ADR-0044.** That citation does not hold — 0044 is about the content map and double reads, not about untransformed content reaching output. Noted in ADR-0207 so the next reader does not go looking.

### Scope note

Sanitisation is **not** a security boundary for a reading agent's judgement. It removes the mechanisms that let repo text impersonate cxpak's own structure. A model that follows instructions found in quoted file content is not addressed here and cannot be addressed by a renderer.

### DIFF
--- a/docs/adrs/0207-repo-text-is-untrusted-at-the-point-it-becomes-output.md
+++ b/docs/adrs/0207-repo-text-is-untrusted-at-the-point-it-becomes-output.md
@@ -1,0 +1,96 @@
+# ADR-0207: Repo text is untrusted at the point it becomes output
+
+- **Status:** ACCEPTED
+- **Date:** 2026-09-03
+- **Issue:** #43
+
+## Context
+
+cxpak reads a repository and hands the result to a reading agent. Everything it
+emits — a commit message, a file body, a symbol body, a diff hunk, a path — is
+text somebody else wrote, and the renderer is where that text stops being data
+and starts being structure.
+
+Two of the three renderers already knew this and one did not:
+
+| renderer | before |
+|---|---|
+| `output/json.rs` | safe by construction — `serde_json` escapes control characters |
+| `output/xml.rs` | `escape_xml` filtered `0x0..=0x8`, `0xB..=0xC`, `0xE..=0x1F` |
+| `output/markdown.rs` | nothing |
+
+Markdown is the default. So the renderer with no control-character policy was
+the one almost every consumer actually reads, and the disagreement between two
+renderers over the same `OutputSections` was invisible because no test compared
+them.
+
+Separately, `code_fence_for` — a function whose entire purpose is to emit a
+backtick run longer than any run inside the content — was used for exactly one
+section, `directory_tree`. The six sites that pack a **file body**, a **symbol
+body** or a **diff hunk** into markdown each wrote `` ``` `` by hand. The
+ordinary consequence is not an attack: `README.md` is a key file, most READMEs
+contain a fenced example, and every one of them closed cxpak's block early and
+had its remainder read as markdown.
+
+## Decision
+
+**Repo-derived text is sanitised where it becomes output, not where it is read.**
+Two mechanisms, one place each.
+
+### 1. `output::strip_control_chars` — one policy, all renderers
+
+Drops every control character except `\n` and `\t`. `escape_xml` now calls it
+instead of carrying its own list, so the two renderers cannot drift again.
+
+`\r` and DEL are dropped where `escape_xml` used to keep them: a lone `\r`
+overwrites a line the reader has already seen, which is the same class of harm
+as `ESC`. A `\r\n` line ending loses its `\r` and keeps its `\n`.
+
+**This is confinement, not censoring.** `\u{1b}[31m` becomes `[31m` — the text
+survives and stops being a terminal instruction. Removing content would trade a
+legibility bug for a fidelity bug.
+
+### 2. `output::fenced` — the fence is sized by the content it must close
+
+`code_fence_for` moves to `output/mod.rs` and all six packing sites go through
+`fenced(content, info)`. In `render_key_files` the fence is computed from the
+**whole** file and reused for whatever survives truncation: dropping lines can
+only remove backtick runs, so a fence that closes the full body closes every
+prefix of it, and the header/footer token accounting stays exact.
+
+### 3. Rejected: fencing whole sections
+
+#43 asks for "fencing on packed content", and for the section bodies that is the
+wrong shape. `metadata`, `module_map`, `git_context` and the rest are **cxpak's
+own generated markdown** — `###` sub-headers, list items, tables. Fencing them
+would turn cxpak's structure into a wall of literal text and destroy the thing
+the format is for.
+
+The vector in those sections is not the section, it is the **interpolation
+point**: `render_git_context` formats a commit message into a single-line list
+item, so a message containing `\n\n## ` breaks out of the item and forges a
+heading. That is fixed at the six interpolation points with `one_line`, which
+flattens newlines and drops backticks for the two fields that sit inside a code
+span. Fencing is used where the payload really is an opaque body, and nowhere
+else.
+
+## Consequences
+
+- One control-character policy, named once, shared by both text renderers.
+- A README with a code block renders correctly in Key Files for the first time.
+- Sanitisation is **not** a security boundary for a reading agent's judgement —
+  it removes the mechanisms that let repo text impersonate cxpak's own
+  structure. A model that follows instructions found in quoted file content is
+  not addressed here, and cannot be addressed by a renderer.
+- `#43` also cites ADR-0044. That citation does not hold: 0044 is about the
+  content map and double reads, not about untransformed content reaching output.
+  Recorded here so the next reader does not go looking.
+
+## Alternatives considered
+
+- **Escape markdown metacharacters** (`#`, `` ` ``, `|`, `*`). Rejected: the
+  section bodies *are* markdown, so escaping them destroys the output, and for
+  the fenced bodies a correctly-sized fence already covers it.
+- **Strip control characters at index time.** Rejected: the index is also the
+  input to search, symbol extraction and the LSP surface, where the file's real
+  bytes are what a consumer wants. Output is the boundary; the index is not.
--- a/docs/adrs/INDEX.md
+++ b/docs/adrs/INDEX.md
@@ -207,3 +207,4 @@
 | [0201](0201-mcp-tool-call-status-discriminator.md) | Machine-readable MCP tool-call status discriminator — Building carries `structuredContent{status,retryable}`, Failed→`isError:true` | ACCEPTED | 2026-07-16 |
 | [0202](0202-graph-node-enumeration-and-seed-validation.md) | Graph query — `nodes` enumeration op + `subgraph` unknown-seed partition + documented id format | ACCEPTED | 2026-07-16 |
 | [0203](0203-self-package-import-resolution-deferred.md) | Self-package import resolution (`use <own-crate>::`) — deferred, limitation documented | ACCEPTED | 2026-07-16 |
+| [0207](0207-repo-text-is-untrusted-at-the-point-it-becomes-output.md) | Repo text is untrusted at the point it becomes output — one control-char policy across renderers, content-sized fences at all six packing sites | ACCEPTED | 2026-09-03 |
--- a/src/commands/diff.rs
+++ b/src/commands/diff.rs
@@ -325,10 +325,8 @@ pub fn run(
     let render_start = std::time::Instant::now();
     let mut full_diff = String::new();
     for change in &changes {
-        full_diff.push_str(&format!(
-            "### {}\n\n```diff\n{}\n```\n\n",
-            change.path, change.diff_text
-        ));
+        full_diff.push_str(&format!("### {}\n\n", change.path));
+        full_diff.push_str(&crate::output::fenced(&change.diff_text, "diff"));
     }
 
     // 9. Budget: diff first, then context signatures with the remainder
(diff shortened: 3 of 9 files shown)


lyubomir-bozhinov:
## Reviewer — **MERGE.** Reviewed at the PR head. CI green. No Blocking, no Important.

An injection surface into the context an agent trusts, so I attacked the guard rather than read it.

### Four mutations

| Mutation | Killed |
|---|---|
| `code_fence_for` ignores CommonMark's up-to-3-space indent | `an_indented_fence_still_counts_up_to_three_spaces` |
| `code_fence_for` always returns ` ``` ` | 4 tests, incl. `content_carrying_a_fence_cannot_close_the_block` |
| `strip_control_chars` strips nothing | 4 tests, incl. `an_ansi_escape_in_repo_text_does_not_reach_markdown_output` |
| **a seventh hand-written fence planted in `trace.rs`** | `no_command_writes_a_markdown_fence_by_hand` |

The fourth is the one I care about. A completeness guard that only fails when you revert the six
known sites is not a completeness guard, so I planted a *new* offender the way a future PR would:

```
a fence written by hand cannot be closed by the content it wraps — use `crate::output::fenced`:
  trace.rs:29: let _planted = format!("```\n{}\n```", "repo text");
```

File, line, offending text, and the remedy. It fires on the seventh site, which is the claim the
module doc makes — *"Fixing the six is not the same as stopping a seventh"* — and the denominator
comes from `read_dir` over `src/commands`, not from a list of known call sites. That is the right
side of the completeness/drift distinction, and `production_lines` counting brace depth rather than
cutting at the first `mod tests` means code *after* a test module is still scanned.

The indent handling is the detail I expected to be wrong and is not. CommonMark lets a closing fence
sit under up to three spaces, so counting only column 0 leaves `   ``` ` able to close the block —
and counting *any* indentation inflates the fence for ordinary indented code. Both directions are
pinned, including four spaces (an indented code block, closes nothing) and `/// ```` ` in a doc
comment. `the_fence_outgrows_the_longest_run_at_a_line_start` also pins that mid-line backticks must
*not* inflate the fence, which is the assertion that stops "just emit twenty backticks" from passing.

### Nice-to-have — the ADR does not say where the sanitiser's edge is, and it has one

Measured what actually survives `strip_control_chars`:

```
ESC / ANSI  a\u{1b}[31m…     stripped        LF, TAB                    kept  (correct)
CR, DEL, NEL                 stripped        U+202E RLO, U+202D LRO     KEPT
                                             U+200B ZWSP, U+2028 LS     KEPT
```

C0 and DEL are handled, which meets #43's acceptance criterion exactly — *"a crafted commit message
/ filename cannot inject headers or ANSI"*. The residue is Unicode: `char::is_control()` is false
for the Cf category, so a bidi override in a filename or commit message reaches the briefing and
renders reversed for a human reading the terminal (Trojan Source). Neither the ADR nor `SECURITY.md`
mentions Unicode at all, so a reader has no way to tell this was scoped out rather than missed.

I am **not** asking for a code change. Stripping all of Cf breaks legitimate content, the fence maths
is unaffected (CommonMark's line endings are `\n`, `\r\n`, `\r` — U+2028 is not one, so it cannot
manufacture a line start your counter missed), and the deception is aimed at a human reader rather
than the model, which sees the codepoint. What would close it is a sentence in the ADR saying C0 and
DEL are in scope and Unicode direction/format characters are not, and why. That is the difference
between a known edge and an unknown one.

Putting the strip policy in one place is the right structural call, and the doc gives the right
reason: `escape_xml` had been filtering C0 since it was written while the markdown renderer had not,
and that asymmetry *is* #43. One helper means they cannot drift apart again — which is the same
argument as the `fenced` helper, one layer up.

lyubomir-bozhinov:
## Reviewer — correcting my own Nice-to-have above. It was an over-claim.

I wrote that a bidi override "reaches the briefing" and pointed at the ADR and `SECURITY.md` not
mentioning Unicode, which reads as *cxpak has no bidi defence*. That is wrong, and I did not check
before saying it.

`src/util.rs` has `sanitize_bidi`, which replaces U+202A–202E and U+2066–2069 with visible
`<U+XXXX>` escapes, and `tests/bidi_sanitization.rs` is a dedicated adversarial suite whose module
doc names the Trojan-source attack explicitly, down to the
`accessLevel = "user\u{202E}//\u{202D}admin"` example. This was thought about carefully and before
I arrived.

**What I actually measured, and all I was entitled to conclude:** `strip_control_chars` — the helper
this PR introduces — does not strip Cf. That is true, and it is narrower than what I said.

Re-derived properly, the call sites are:

```
sanitize_bidi        recent_changes.rs:56  ·  lsp/methods.rs:268,307  ·  architecture.rs:213,221
                     visual/search_index.rs:50,74,75,104          (paths, symbols, prefixes)

markdown.rs:13,59    strip_control_chars only
```

So the accurate finding is narrower and, I think, more useful than the one I posted: **the markdown
renderer is the one output path that does not call `sanitize_bidi`.** The visual and LSP surfaces
escape bidi in the strings they consume; the markdown briefing strips C0 and passes Cf through, on
file bodies and on git text.

That is the same shape as #43's own thesis — *"The XML renderer was hardened and the markdown one
was not, which is the whole asymmetry"* — recurring one layer up with a different character class.
Which is an argument for this PR's structural choice rather than against it: one helper is what
stops the *next* asymmetry, and `strip_control_chars` is now the single place a `sanitize_bidi` call
would go.

Nothing here changes the merge; the fix is right and the mutations stand. Filing the residue
separately rather than leaving it in a merged PR's comments, since it is a decision about scope
(a briefing that escapes bidi in symbol names and not in the code around them is defensible — the
model sees the codepoint either way — but it should be a decision, not an accident of which
renderer got hardened when).

PR #129, opened 2026-09-03
fix(#62): the visual ceiling is a context budget, not a transport limit

### LINKED ISSUE #62: cxpak_insight op=visual returns a 340KB HTML dashboard (90% inlined d3) into the caller's context, with no budget parameter
## Summary

`cxpak_insight` with `op: "visual"` returns a **340,432-byte** self-contained HTML dashboard as the
MCP tool result — directly into the caller's context window. Measured on a **three-file** repository.

```
total bytes:        340,432      (~85,000 tokens at 4 bytes/token)
  5 <script> blocks 308,254      largest single block: 279,703  (inlined d3)
  1 <style> block    31,775
```

Roughly 90% of the payload is an inlined JavaScript library. The actual repository content is a
rounding error.

## Its siblings are three orders of magnitude smaller

Same server, same repo, same session — every other `cxpak_insight` op:

| op | bytes |
|---|---|
| `drift` | 57 |
| `health` | 154 |
| `security_surface` | 447 |
| `risks` | 744 |
| `architecture` | 1,009 |
| `onboard` | 1,102 |
| `conventions` | 3,916 |
| **`visual`** | **340,432** |

## The consumer cannot bound it, because the parameter does not exist

`tools/list` advertises `cxpak_insight` with exactly one property:

```json
{ "properties": { "op": {...} }, "required": ["op"] }
```

No `tokens`, no `format`, no `out`. The description is
*"Report health, risk, and architecture insights. Select a capability via `op` ∈ {health, risks,
architecture, conventions, security_surface, drift, visual, onboard}"* — `visual` sits in that list
beside seven ops that return about a kilobyte, with nothing marking it as different.

An agent choosing an op from that list will eventually choose this one, and spend most of a context
window receiving a d3 bundle.

## The CLI shows what this command is actually for

```
cxpak visual [OPTIONS] [PATH]
  --visual-type <VISUAL_TYPE>  all | dashboard | architecture | risk | flow | timeline | diff
  --format <FORMAT>            html | mermaid | svg | png | c4 | json | cypher | graphml
  --out <OUT>
```

`--out` and a `png`/`svg` format make the intent plain: **write a dashboard to a file and open it.**
There is no `--tokens` on this command — correctly, because a file is not a context window.

The defect is the wiring, not the command: a file-producing generator was exposed on a transport
whose only return channel is the model's context, and the MCP surface exposes none of `--out`,
`--format` or `--visual-type` that would make it viable there.

This also sits oddly against the project's own framing — *"Spends CPU cycles so you don't spend
tokens"* — since this is the one surface that spends tokens without a budget.

## Reproduction

```sh
mkdir r && cd r && git init -q . && echo 'def a(): return 1' > a.py && git add -A && git commit -qm init
{ echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"qa","version":"1"}}}'
  echo '{"jsonrpc":"2.0","method":"notifications/initialized"}'; sleep 4
  echo '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"cxpak_insight","arguments":{"op":"visual"}}}'
  sleep 10
} | docker run --rm -i -v "$(pwd):/repo" ghcr.io/barnett-studios/cxpak:latest serve --mcp . | wc -c
```

## Provenance

`Provenance: pre-existing — `visual` has been in the `cxpak_insight` op set since the tool was
consolidated; nothing about the payload size is new, and no budget parameter has ever existed on
this tool.`

## Last known good

`never observed green` — there is no version of the MCP surface where this op returns a
context-sized payload.

## Suggested fix

Pick whichever matches the intent:

1. **Write it, return the path.** Mirror the CLI: generate to a path under the repo (or a
   caller-supplied `out`) and return `{"path": "...", "bytes": N}`. The model gets a pointer, the
   human opens the file.
2. **Return a small format over MCP.** `mermaid` or `json` are already supported by the CLI's
   `--format`; either is context-appropriate. HTML is the one that is not.
3. **Or drop `visual` from the MCP op set entirely** and leave it a CLI command. That is the
   smallest change and loses nothing an agent can use.

Whichever is chosen, `cxpak_insight` should accept a `tokens` budget like the context surface does,
so no op can exceed what the caller asked for.

## Verified correct in the same pass (recorded so nobody re-checks)

The graph surface is accurate on a repo with known ground truth
(`app.main → helper.compute → util.add`, two uncalled functions):

- `cxpak_graph op=call_graph` → exactly 2 edges, both `confidence: "Exact"`, `unresolved: []` ✓
- `cxpak_graph op=dead_code` → exactly `util.unused_util` and `helper.orphan`, `total: 2`; correctly
  did **not** flag the entry point or either live function ✓
- `cxpak_graph op=api_surface` → `total: 5`, matching the five defined symbols ✓
- `graph`/`trace`/`blast_radius`/`predict` correctly reject missing required arguments
  (`id`, `target`, `files`) with `isError: true` ✓

## Related

- cxpak#45 — MCP wire contract: no per-op schema/version. The single-property `cxpak_insight` schema
  here is the same gap seen from the payload side.
- cxpak#32 — `visual` compiled into the default feature set.

Found by the QA loop, pass 9.

lyubomir-bozhinov:
Closes #62.

**Provenance: `never-worked`** — as the issue records. `visual` has been in the `cxpak_insight` op set since the tool was consolidated, and no version of this surface has ever returned a context-sized payload. So there is no gate that should have fired, only a threshold that was never in the right units.

---

## The guard was there. It was measuring the wrong thing.

```rust
const MCP_INLINE_LIMIT: usize = 1_048_576; // 1 MB
if format == "html" && content.len() > MCP_INLINE_LIMIT {
```

[ADR-0135](../blob/main/docs/adrs/0135-mcp-html-inline-1mb-spill-to-disk.md) names the hazard correctly — *"can blow **context/transport** limits"* — and then picks a number for the second word. 1 MB is a transport figure; a context window is three orders of magnitude smaller in the units that matter.

The measured payload is **340,432 bytes from a three-file repository**, ~90% of it an inlined d3 bundle. That is **32× under** the ceiling, so the guard ran, reported nothing, and did not do its job — sitting in a menu beside seven ops that return 57 to 3,916 bytes.

Two faults, one cause:

| | |
|---|---|
| the limit came from the wrong **domain** | bytes of transport, not tokens of context |
| `format == "html"` covered the wrong **set** | one of eight formats; `graphml`, `json` and `cypher` over a large graph are not small, and none of them was bounded at all |

ADR-0135 lists that second one under **Neutral**. It is not neutral.

**Its revisit conditions could not fire.** They are *"MCP transport limits change"* and *"inline vs path handling confuses tool consumers"*. Neither can describe a payload **under** the threshold exhausting a context window — the guard's own trigger conditions were blind to its own failure mode.

## What changes

- **`MAX_MCP_VISUAL_TOKENS` = `MAX_MCP_CONVENTIONS_TOKENS` (5,000).** Not a new number, deliberately: `visual` is one entry in a menu of eight, and an agent picking it off that list should not pay two orders of magnitude more than its neighbours cost. Pinned by an assertion so the two cannot drift.
- **`tokens` is honoured**, parsed exactly as `conventions` parses it. Before this the op had **no caller lever at all**, which is half of what the issue is about.
- **The spill covers every format**, and the file carries the extension of what was written — a mermaid graph saved as `.html` is a file nothing will open.
- **The spill result is JSON** — `{path, bytes, tokens, token_budget, format, note}` — because the caller has to decide what to do next and a prose sentence makes it guess.

**Spill, not truncate.** Every format here is structured, and half a structured document is not a smaller document, it is a broken one. `conventions` truncates because its payload degrades gracefully; this one writes the whole artifact and returns where it went — which is also what the CLI's `--out` says the command is for.

**Kept unchanged from 0135:** the spill mechanism, `.cxpak/visual/` as the destination, and the path-escape check.

---

## Adversarial self-check

**Mutations: 5/5 killed.**

| | mutation | killed by |
|---|---|---|
| V1 | the pre-fix 1 MB byte ceiling, html only | `an_over_budget_visual_is_written_to_disk_not_returned` |
| V2 | `tokens` read but ignored | `a_larger_tokens_budget_returns_the_payload_inline` |
| V3 | **CONTROL** — spill *everything* | `a_within_budget_visual_is_still_returned_inline` |
| V4 | mermaid spills as `.html` | `every_visual_format_spills_under_its_own_extension` |
| V5 | the budget drifts from `conventions`' cap | `the_visual_budget_matches_the_other_capped_op` |

**V3 is the control that matters.** "Do not return the payload" is trivially satisfied by never returning it — which passes V1's test and destroys the op. The inline test fails if the fix takes that route.

The format→extension table is **hand-written from the CLI's documented `--format` set**, not read back from the function under test, so it states what the mapping must be rather than confirming the code agrees with itself.

### Two guardians caught the behaviour change, and both were right

**1. `mcp_visual_dashboard_returns_html`** (live MCP server, end to end) already accepted either branch — but detected the spill by matching the prose `"Output written to"`, which this replaces with JSON. It failed. Now asserts the **contract**: a pointer must parse, name a file that really exists, report that file's real byte count, and report tokens over the budget it names.

`grep -rn "Output written to"` over `src`, `tests`, `docs` and `README.md` returns only that test's two lines, so that is the whole radius for the message shape.

**2. `invariant_mcp_inline_limit_constant_present`** is a grep over `serve.rs` pinning ADR-0135's constant. Removing the constant made it red — exactly what a pinned invariant is for.

**Repointed, not deleted.** The mechanism it guards survived; only the number and the units moved, and deleting the guard because the constant was renamed would retire a real invariant on a technicality. It now pins `MAX_MCP_VISUAL_TOKENS` and `visual_format_extension`, and asserts the byte threshold does **not** come back beside the token one — two ceilings on one payload is how they drift. It stays a grep for the reason it always was: it catches accidental *removal*, which no behavioural test can, since a behavioural test for "a limit exists" passes on any limit including 1 MB.

**The sibling formats are the control for the widened condition.** `mcp_visual_architecture_returns_mermaid_with_graph_header`, `_c4_contains_workspace`, `_json_parseable_content` and `_svg_is_non_empty` all still pass **inline**. Those formats are now bounded too and did not become "always writes a file".

### What I did NOT verify

- **No live MCP server was driven against a large real repository.** The over-budget case is exercised through the in-process handler and through the spawned-server integration test on the fixture repo; I did not reproduce the issue's original 340 KB measurement end to end.
- **Whether any consumer depends on the inline HTML.** This is a behaviour change on a published MCP surface — that is the point, and it is recorded as a negative consequence in ADR-0208, but I did not survey consumers.
- **The rendering cost itself.** The token count is computed over the fully rendered payload, so a large dashboard is still built before being written. This bounds what the caller *receives*, not what the server *does*.
- **Windows/Linux path behaviour** of the spill. Unchanged from 0135, and untested there before or after.

### Not in scope

`tools/list` still advertises only `op`, so `tokens` is honoured but **not discoverable**. `additionalProperties: true` and per-op parameter schemas are **cxpak#45**, which #62 itself names as the same gap seen from the other side. Recorded in ADR-0208's revisit conditions rather than half-fixed here.

### Docs

- **ADR-0208** supersedes **ADR-0135**, which is marked `SUPERSEDED` with a pointer rather than left to contradict the code.
- `docs/MIGRATION-3.0.md` line 81 enumerated this op's params and now names `tokens` and the spill contract (`docs-with-code`).
- ADR numbering: 0206 is taken by PR #127 (issue #40) and 0207 by PR #128 (issue #43), both open. `INDEX.md` will conflict on adjacent lines if those land in a different order — see **cxpak#126**, where 0204 and 0205 are on disk and absent from the index, which is why the high-water mark cannot simply be read off it.

### DIFF
--- a/docs/MIGRATION-3.0.md
+++ b/docs/MIGRATION-3.0.md
@@ -83,7 +83,7 @@ be **removed in a future release**. Migrate to the intent-tool + `op` form.
 | `cxpak_security_surface` | `cxpak_insight` | `security_surface` | add `op`; params unchanged (`focus`) |
 | `cxpak_data_flow` | `cxpak_graph` | `data_flow` | add `op`; params unchanged (`symbol`, `sink`, `depth`, `focus`) |
 | `cxpak_cross_lang` | `cxpak_graph` | `cross_lang` | add `op`; params unchanged (`file`, `focus`) |
-| `cxpak_visual` | `cxpak_insight` | `visual` | add `op`; params unchanged (`type`, `format`, `focus`, `symbol`, `files`) |
+| `cxpak_visual` | `cxpak_insight` | `visual` | add `op`; params `type`, `format`, `focus`, `symbol`, `files` unchanged, **plus `tokens`** (cxpak#62) — a payload over the budget is written under `.cxpak/visual/` and the call returns `{path, bytes, tokens, token_budget, format}` instead of the artifact. Default 5 000 tokens, the same ceiling `conventions` uses |
 | `cxpak_onboard` | `cxpak_insight` | `onboard` | add `op`; params unchanged (`focus`, `format`) |
 
 ### New / re-keyed ops with a sub-selector
--- a/docs/adrs/0135-mcp-html-inline-1mb-spill-to-disk.md
+++ b/docs/adrs/0135-mcp-html-inline-1mb-spill-to-disk.md
@@ -1,7 +1,7 @@
 ---
 id: '0135'
 title: Spill MCP visual HTML responses over 1MB to disk, return file path
-status: ACCEPTED
+status: SUPERSEDED by ADR-0208
 date: 2026-04-01
 triggered_by: cxpak_visual MCP tool response sizing (Task 15)
 loop: implementation
@@ -10,6 +10,11 @@ loop: implementation
 # ADR-0135: Spill MCP visual HTML responses over 1MB to disk, return file path
 
 ## Context
+
+> **SUPERSEDED by [ADR-0208](0208-the-mcp-visual-ceiling-is-a-context-budget.md) (cxpak#62).**
+> The spill mechanism below is kept. The 1 MB threshold and the `format == "html"` gate are not:
+> the limit was a transport figure applied to a context problem, and a 340 KB dashboard from a
+> three-file repository passed under it. Read 0208 for what runs today.
 
 Shipped in v2.0.0. Returning large self-contained HTML inline through the MCP tool channel can blow context/transport limits. The `cxpak_visual` tool (Task 15) needed a threshold strategy to keep payloads bounded.
 
--- a/docs/adrs/0208-the-mcp-visual-ceiling-is-a-context-budget.md
+++ b/docs/adrs/0208-the-mcp-visual-ceiling-is-a-context-budget.md
@@ -1,0 +1,100 @@
+---
+id: '0208'
+title: The MCP `visual` ceiling is a context budget, not a transport limit
+status: ACCEPTED
+date: 2026-09-03
+triggered_by: cxpak#62 — 340 KB dashboard returned inline from a three-file repository
+loop: implementation
+supersedes: '0135'
+---
+
+# ADR-0208: The MCP `visual` ceiling is a context budget, not a transport limit
+
+## Context
+
+Supersedes [ADR-0135](0135-mcp-html-inline-1mb-spill-to-disk.md), which shipped the
+spill-to-disk mechanism this keeps and the threshold this replaces.
+
+0135 opens by naming the right hazard:
+
+> Returning large self-contained HTML inline through the MCP tool channel can blow
+> **context/transport** limits.
+
+Then it picks a number for the second word. `MCP_INLINE_LIMIT = 1_048_576` is a transport
+figure; a context window is three orders of magnitude smaller in the units that matter.
+
+cxpak#62 measured the consequence on a **three-file** repository:
+
+| op | bytes |
+|---|---|
+| `drift` | 57 |
+| `health` | 154 |
+| `security_surface` | 447 |
+| `risks` | 744 |
+| `architecture` | 1,009 |
+| `onboard` | 1,102 |
+| `conventions` | 3,916 |
+| **`visual`** | **340,432** |
+
+~90% of that payload is an inlined d3 bundle. It is 32× under 0135's ceiling, so the guard
+was working exactly as designed and the caller lost most of a context window anyway.
+
+0135 also lists this under **Neutral**:
+
+> The spill check is gated strictly on `format == "html"`.
+
+It is not neutral. Seven other formats are reachable through the same op, and `graphml`,
+`json` and `cypher` over a large graph are not small. None of them was bounded at all.
+
+**Its revisit conditions could not fire.** They are *"MCP transport limits change"* and
+*"inline vs path handling confuses tool consumers"*. The failure that happened is neither: a
+payload well **under** the threshold exhausting a context window is invisible to both. A
+guard whose trigger conditions cannot describe its own failure mode is the shape ADR-0060 is
+about, one layer out.
+
+## Decision
+
+**The ceiling is measured in tokens, against a caller-supplied budget, for every format.**
+
+- `MAX_MCP_VISUAL_TOKENS` = `MAX_MCP_CONVENTIONS_TOKENS` (5,000). Not a new number: `visual`
+  is one entry in a menu of eight ops, and an agent picking it off that list should not pay
+  two orders of magnitude more than its neighbours cost.
+- `tokens` is honoured, parsed the same way `conventions` parses it. A caller that wants the
+  whole artifact inline says so. Before this there was no such lever on this op.
+- The spill applies to **all** formats, and the file gets the extension of what was written —
+  a mermaid graph saved as `.html` is a file nothing will open.
+- The spill result is JSON — `{path, bytes, tokens, token_budget, format, note}` — because
+  the caller has to decide what to do next and a prose sentence makes it guess.
+
+**Spill, not truncate.** Every format here is structured, and half a structured document is
+not a smaller document, it is a broken one. `conventions` truncates because its payload
+degrades gracefully; this one writes the whole artifact and returns where it went. That also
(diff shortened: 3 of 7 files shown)


lyubomir-bozhinov:
**Not reviewed — the branch conflicts with `main`.** #123, #125, #128, #130 and #121 landed today and this touches `serve.rs` and `docs/adrs/INDEX.md`, both of which moved.

`gh` reports `CONFLICTING / DIRTY`. Rebasing is yours; I review, I do not push to a PR's branch. #127 is in the same state.

I have not started the review, so nothing here is a verdict — this is only to say why it is sitting. Ping me when the head moves and I will take it fresh.

lyubomir-bozhinov:
Rebased onto `ebbbfa4` — head is now `33ba423`, `MERGEABLE`. Ready for a fresh look.

One conflict, in `docs/adrs/INDEX.md` only. `src/commands/serve.rs` and `docs/MIGRATION-3.0.md` auto-merged. #123, #125, #127, #128, #130 and #121 had all appended index rows while this sat, so the resolution was an append: main's `0206` and `0207` rows kept, my `0208` row after them. No row was dropped and no marker survived.

Re-verified after the rebase rather than assuming the merge was clean:

```
cargo test --lib                                  2678 passed; 0 failed; 3 ignored
cargo test mcp_visual                             4 passed  (both repointed guardians)
cargo clippy --workspace --all-targets -D warnings  clean
```

**Worth flagging because it bit me during the resolution: #126 is live and I hit it.** `0204` and `0205` are on disk with no `INDEX.md` row, so the index reads `0203` as the high-water mark while the directory reads `0205`. Resolving an append conflict in that file means deciding the next free number, and the index is the wrong oracle for it — I allocated `0208` by listing `docs/adrs/` and checking every remote branch, not by reading the index. Had I trusted the index I would have picked a colliding number and the conflict would have resolved into a real defect rather than out of one.

I have not touched that drift here: it is pre-existing, it is already filed as #126, and fixing it inside an unrelated PR would hide one change's risk behind another's review. Recording it because #126 currently reads as a tidiness issue and this is the shape of it doing damage.

**Not verified:** the full integration suite after the rebase — I ran the lib suite, the targeted guardians and clippy, and left the rest to CI.

Issue #81, opened 2026-08-28
Four capabilities declare a focus parameter their handlers never read — health declares nothing else, and review ignores both of its parameters


lyubomir-bozhinov:
**Severity: Medium** — measured on published 3.1.4, and the code is unchanged on `main`.

`src/capability/mod.rs` declares each capability's parameter set. **27 of the 29 capabilities declare
`focus`.** Four of those four-and-twenty-seven have MCP handlers that never read it: the argument is
accepted, silently discarded, and the caller gets the whole-repo answer to a question it scoped.

```
health       params=["focus"]                          s(mcp, lsp, http, visual)
drift        params=["save_baseline", "focus"]         s(mcp, lsp, http)
review       params=["since", "focus"]                 s(mcp)
data_flow    params=["symbol", "sink", "depth", "focus"]  s(mcp, lsp, http)
```

`health` is the sharp one: `focus` is its **only** declared parameter, and it is projected on four
surfaces.

`review` is worse than the heading says — **neither** of its two declared parameters is read.
`grep -c '"since"' src/commands/serve.rs` returns **0**. Controls for that grep, same file, same
pattern shape: `"limit"` → 9, `"focus"` → 35, `"tokens"` → 37. The pattern works; `since` is simply
not there.

## Measured

Three calls per op on a three-file fixture (`widget.py` defines `Widget`; `app.py` imports and calls
it; `leaf.py` is isolated), published 3.1.4 over MCP stdio:

```
op                  {op}     {op, focus:"widget.py"}   {op, focus:"nosuchdir/"}
insight/health      155 B    155 B  (identical)        155 B  (identical)
insight/drift       248 B    248 B  (identical)        248 B  (identical)
review/review       242 B    242 B  (identical)        242 B  (identical)
```

Byte-identical across a prefix that matches one of three files and a prefix that matches none. The
control that makes this a result rather than "focus never narrows anything on a small fixture" is on
the same server in the same session — ops whose handlers **do** read it change:

```
graph/api_surface   1012 B (app, leaf, widget)   537 B (widget only)   177 B (none)
insight/risks        714 B (app, leaf, widget)   261 B (widget only)     2 B (none)
insight/architecture 823 B (app, leaf, widget)   243 B (widget only)    42 B (none)
```

## Confirmed in the source

```rust
"health" => {
    let health = crate::intelligence::health::compute_health(index);   // no `focus`
    …
}
"drift" => {
[...]

PR #130, opened 2026-09-03
fix(#64): the installed plugin's README named a surface that no longer exists

### LINKED ISSUE #64: The installed plugin's README advertises 13 MCP tools; the server advertises 5, and none of the 13 by those names

lyubomir-bozhinov:
Closes #64.

**Provenance: `regression — 3.0 consolidation (ADR-0182)`.** The 13-row table was correct when written; the consolidation removed the surface it describes and did not carry the documentation with it. The gate that should have fired is `tests/advertised_tool_names.rs`, which existed and scanned `src/` only — this widens it, which is the obligation a regression carries.

---

## What was wrong

`plugin/README.md:51` was headed **"MCP Tools (13)"** and listed the complete v2.x surface. `tools/list` serves `mcp_tool_schemas()`, which renders the intent catalog — **five** tools, each selecting a capability via `op`. **Not one** of the 13 names appears in the advertised set.

That file is the README of the plugin a user *installs*. It is what they read to learn what they just got.

All 13 still **route**, through the 26-arm alias table, so a permissive client that copies a name out of the README succeeds and nothing fails loudly. A client that builds its callable set from `tools/list` — which is what an agent harness does — gets none of them. The doc was wrong in the one direction the alias hides, which is why it survived.

### The sibling the ticket did not count

`README.md:187` names `cxpak_auto_context` and `cxpak_health` too. **There it is legitimate** — that paragraph's subject *is* the deprecated names — but the sentence around them was not. It said the aliases remain callable *"for one release"*, a horizon that has passed and that nothing enforced.

Per the acceptance criterion, I **dropped the claim** rather than inventing a date: setting a removal release is the owner's call, and this repo's own downstream `CLAUDE.md` already records the aliases as live-until-further-notice. It now says they route, that `tools/list` does not advertise them, and that no removal release is set.

### The `focus` claim, measured rather than carried over

The old section said *"all tools support a `focus` path prefix parameter"*. Measured on `main`:

```
23 of 29 ops read args.get("focus")
6 do not:  data_flow  drift  graph  health  retrieval  review
```

So the new section makes **no blanket parameter claim at all**, and says why in the document: a sentence that is right for most ops is wrong for the caller holding one of the others.

That measurement also turned up a **fifth** instance of **cxpak#81** — `visual` declares `focus` at `src/capability/mod.rs:492` and discards it at `src/commands/serve.rs:4326` with `let _ = focus`. **Reported on #81, not fixed here**, along with the caveat that the 23/29 figure counts `visual` as reading `focus` (it does; the discard is one line later), so that number is a floor rather than a total.

## The guard, which is the actual fix

`tests/advertised_tool_names.rs` already scanned `src/` for `cxpak_*` names in string literals — **the same defect, one scope too narrow.** It now also scans markdown.

**Denominator: the whole tree.** A new document is covered the day it is added, not the day someone remembers to list it — mutation D2 is exactly that case. Two escapes, and the difference between them is deliberate:

- **`docs/adrs/` is excluded by rule.** An ADR is a dated record of a decision taken when those names existed; rewriting one to use today's names would falsify history. There is no version of this guard an ADR should pass.
- **Everything else opts out with `advertised-tool-names: exempt` in the file**, beside the text it excuses. Two documents legitimately name the dead surface — the migration table, whose subject is the mapping, and README's deprecation paragraph. Declaring it in the document means deleting the paragraph deletes the exemption; a path list in the test would outlive its reason.

The expectation is deliberately **shared** with the thing it checks — `mcp_catalog_tools()` is the function `tools/list` is built from, so a rename moves both sides together. This is a *drift* claim, and drift needs a shared denominator. (A *completeness* claim needs the opposite, which is why the sibling test over `src/` derives its site set from the source rather than from a list.)

---

## Adversarial self-check

**Mutations: 5/5 killed.**

| | mutation | killed by |
|---|---|---|
| D1 | the plugin README names a dead tool again | the violation list, by name |
| D2 | a **new** doc names a dead tool | the violation list — the case a path allowlist misses |
| D3 | **CONTROL** — both scanned docs exempt themselves | `checked > 0` |
| D4 | **CONTROL** — the catalog is empty | fails loudly rather than passing |
| D5 | the ADR rule widens to swallow `docs/` + `plugin/` | `checked > 0` |

**D3 and D5 corrected a comment I had written before measuring it.** I predicted `scanned >= 2` would catch them. It does not: ~17 other markdown files in the tree mention no tool at all, so the file count stays healthy while the guard has stopped checking anything. `checked > 0` is the control that bears the weight, and the comment now says so — with `scanned` kept for the one thing it catches that `checked` cannot, a moved or renamed root.

That is the whole reason both controls are there. "Every name in the docs is advertised" is trivially satisfied by scanning nothing.

### What I did NOT verify

- **The five tools' `op` lists in the new table are transcribed from `mcp_catalog_tools()` at this commit, and are not asserted by a test.** The guard checks tool *names*, not the op enumerations beside them. A future op added to a tool leaves that table stale and nothing here catches it — the sibling `src/` test does check `(op: "...")` pairs, but only inside Rust string literals. Worth its own guard; not in this change.
- **The plugin is not installed and driven end to end.** I read `mcp_catalog_tools()` through a throwaway example binary (removed) rather than by running `cxpak serve --mcp` against a live client.
- **Whether any published artifact ships the old README.** This fixes the tree; a released plugin package already in someone's `~/.claude` still carries the 13-row table until a release goes out — which is `.github#24`'s subject.
- **The other 22 ops that read `focus`** were not each checked for a `let _ =` or an unused binding. `visual` was found incidentally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,10 @@ Your AI tool gets five intent-parameterized tools; each selects a capability via
 | **`cxpak_review`** | `review`, `diff`, `verify` |
 | **`cxpak_insight`** | `health`, `risks`, `architecture`, `conventions`, `security_surface`, `drift`, `visual`, `onboard` |
 
-The v2.x per-tool names (`cxpak_auto_context`, `cxpak_health`, ...) remain callable as deprecated aliases for one release. See [`docs/MIGRATION-3.0.md`](docs/MIGRATION-3.0.md).
+<!-- advertised-tool-names: exempt — this paragraph's subject IS the pre-3.0 names, so naming
+     them is its job. The exemption is declared here rather than hardcoded in the test, so moving
+     or deleting this paragraph removes the exemption with it. -->
+The v2.x per-tool names (`cxpak_auto_context`, `cxpak_health`, ...) remain callable as deprecated aliases; **`tools/list` does not advertise them**, so a client that builds its callable set from the advertised surface will not find them. No removal release is set. See [`docs/MIGRATION-3.0.md`](docs/MIGRATION-3.0.md).
 
 On large repositories, `cxpak serve --mcp` answers the MCP `initialize` handshake immediately and builds the index in the background -- fixing the startup timeout. Tool calls that arrive before the index is ready get a graceful retry status, then byte-identical results once it is built.
 
--- a/docs/MIGRATION-3.0.md
+++ b/docs/MIGRATION-3.0.md
@@ -1,3 +1,8 @@
+<!-- advertised-tool-names: exempt — this document's SUBJECT is the pre-3.0 tool names and
+     what they map onto. A guard that required every name here to be advertised would require
+     this file to stop being a migration guide. Declared here rather than as a path in the
+     test, so the exemption cannot outlive the document. -->
+
 # Migrating to cxpak 3.0.0 — MCP tool consolidation (BREAKING)
 
 cxpak 3.0.0 replaces the 26 hand-rolled MCP tools with **five intent-parameterized
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -48,25 +48,34 @@ cargo install cxpak
 
 All commands ask for a token budget (default: 50k).
 
-## MCP Tools (13)
+## MCP Tools (5)
 
-When used as an MCP server (`cxpak serve --mcp`), all tools support a `focus` path prefix parameter for scoped results:
+When used as an MCP server (`cxpak serve --mcp`), cxpak advertises **five intent-tools**. Each
+one selects what it does through a required `op` parameter, so the tool you call is the
+*intent* and the `op` is the *capability*:
 
-| Tool | Description |
+| Tool | `op` values |
 |------|-------------|
-| `cxpak_auto_context` | One-call optimal context for any task (includes DNA section) |
-| `cxpak_overview` | Structured repo summary |
-| `cxpak_trace` | Trace a symbol through dependencies |
-| `cxpak_stats` | Language stats and token counts |
-| `cxpak_diff` | Show changes with dependency context |
-| `cxpak_context_for_task` | Score and rank files by relevance to a task |
-| `cxpak_pack_context` | Pack selected files into a token-budgeted bundle |
-| `cxpak_search` | Regex search with context lines |
-| `cxpak_blast_radius` | Analyze change impact with risk scores |
-| `cxpak_api_surface` | Extract public API surface |
-| `cxpak_context_diff` | Show what changed since last auto_context call |
-| `cxpak_verify` | Verify code changes against observed conventions |
-| `cxpak_conventions` | Return the full convention profile with evidence |
+| `cxpak_context` | `context`, `retrieval`, `search`, `overview`, `stats`, `context_for_task`, `pack_context`, `briefing` |
+| `cxpak_graph` | `graph`, `trace`, `blast_radius`, `call_graph`, `dead_code`, `api_surface`, `data_flow`, `cross_lang`, `predict` |
+| `cxpak_data` | `data` |
+| `cxpak_review` | `review`, `diff`, `verify` |
+| `cxpak_insight` | `health`, `risks`, `architecture`, `conventions`, `security_surface`, `drift`, `visual`, `onboard` |
+
+Start with `cxpak_context` (`op: "context"`) — it returns token-budgeted structural context for a
+one-line task description.
+
+Per-op parameters are listed in [`docs/MIGRATION-3.0.md`](../docs/MIGRATION-3.0.md), and
+`tools/list` is the authority at runtime.
+
+This section deliberately makes **no blanket claim about parameters**. The one it replaced said
+*"all tools support a `focus` path prefix parameter"*, and that was not true of the surface it
+described or of this one — measured on `main`, 23 of 29 ops read `focus` and six do not. A
+sentence that is right for most ops is wrong for the caller holding one of the others.
+
+> The pre-3.0 per-tool names still route as deprecated aliases, but `tools/list` does not
+> advertise them — a client that builds its callable set from `tools/list`, which is what an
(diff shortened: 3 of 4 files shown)


lyubomir-bozhinov:

The fix is a documentation rewrite, so the only question that matters is whether anything stops it
regressing. Three mutations, all fire with file, line and the offending name:

```
M1  a dead name re-added to plugin/README.md
      plugin/README.md:84 names "cxpak_pack_context", which tools/list does not advertise
M2  a BRAND NEW doc file (docs/reviewer-probe.md) naming a dead tool
      docs/reviewer-probe.md:3 names "cxpak_blast_radius", which tools/list does not advertise
M3  a dead name in a src string literal
      src/commands/overview.rs:42 names "cxpak_api_surface", which tools/list does not advertise
```

**M2 is the one worth the run.** Reverting the file the ticket names only proves the file was
edited; a *new* document is where the next stale name appears, and the scan covers it the day it
exists. That is the claim the doc-comment makes — *"Every `.md` in the tree is scanned, so a new
document is covered the day it is added"* — and it holds.

Both sides come from `mcp_catalog_tools()`, the same function `tools/list` is built from, so a
rename moves the expectation and the advertisement together. That is the right construction for a
**drift** check, and the opposite of what a completeness check needs — the two halves of this file
get it right in both directions: the *site* set is derived from the tree (completeness), the
*advertised* set is shared with the server (drift).

The module doc also records why the first attempt was insufficient, in the terms that generalise:

> The first fix corrected the two strings the ticket named. That is a hand-written denominator.

Three further sites fell out of re-deriving it — the `context_for_task` hint in `serve.rs`, reached
through an *advertised* entry point, and two `visual` error strings. None were in the ticket. A
ticket under-counting its own defect is the usual direction, and the remedy is exactly this: derive
the denominator, do not transcribe it.

The two exemptions are narrow and correctly chosen — `legacy_alias_to_op`, whose 26 arms exist *to*
name pre-3.0 tools, and the migration table plus README deprecation section, whose subject is the
old names. Each is matched by a marker in the item's own declaration, so renaming the function makes
its arms visible to the scan and fails loudly rather than silently widening it. That is the right
failure direction for an exemption.

lyubomir-bozhinov:
**Follow-up: this broke `coverage` on `main`, and my review is why it got through.** Filed as #132.

`every_tool_name_in_current_docs_is_advertised` walks every `.md` in the tree and does not exclude `.cxpak/`. CI runs cxpak before the `coverage` job, so `.cxpak/modules.md` exists there and the gate reads cxpak's own generated output as documentation — eight failures naming pre-3.0 tools that only appear in the cache. `test` still passes; the two jobs differ in whether cxpak has run first.

My verdict above singled out the tree-wide scan as the strongest thing in the PR, and I planted `docs/reviewer-probe.md` to prove a brand-new document is covered. That was the right probe for one direction and I did not run the other: **the same property that covers a new document covers a generated one**, and the artifact most likely to be generated in this repo is the one this tool writes. Testing the file I invented and not the file the tool invents was the gap.

The finding stands and the fix is narrow — exclude `.cxpak/` in `markdown_files`, do not switch the walk to `git ls-files` (that trades this false positive for the missed-new-document failure the module doc exists to prevent). Details and the local repro in #132.

